PR #16, opened 2021-02-05
Show Download Release File button after successful release

jbeast:


khelwood:
What does this do if you multiple release ids (from releasing multiple labware)?

khelwood:
The download button should include all the release ids in one url, to get a single file containing all the stuff you've just released.

jbeast:
Ah, I've not understood the release API properly. Thought each release operation would have an ID, rather than each piece of released labware. Will fix it! Thanks 👍

khelwood:
FYI in CGAP LIMS everything you release in one go is bundled together in one goal; but that's a pain because when someone subsequently asks me to alter the date or some detail about something that was released, I've got to manually separate it out into separate goals. So in Stan each labware has an individual release.

khelwood (approved):

jbeast:
> FYI in CGAP LIMS everything you release in one go is bundled together in one goal; but that's a pain because when someone subsequently asks me to alter the date or some detail about something that was released, I've got to manually separate it out into separate goals. So in Stan each labware has an individual release.

Yeah, that makes perfect sense. I've fixed the download link in the PR include all the releases now.

PR #27, opened 2021-03-25
Cdna release

khelwood:
Validate bio state on release page

Labware scan panels now support a check function for labware that is run before
the labware is added to the table.
The release page uses this to check labware's bio state.
Labware scan panel columns relating to samples now show info for all samples in the labware, not just the first.
For the release page (and probably others) it would probably be better if the table had a row per
labware-slot-sample instead of a row per labware.

### DIFF
--- a/src/components/labwareScanPanel/columns.tsx
+++ b/src/components/labwareScanPanel/columns.tsx
@@ -1,8 +1,8 @@
 import React from "react";
-import { CellProps, Column } from "react-table";
-import { Labware, LabwareLayoutFragment } from "../../types/graphql";
+import {CellProps, Column} from "react-table";
+import {Labware, LabwareLayoutFragment} from "../../types/graphql";
 import LabelPrinterButton from "../LabelPrinterButton";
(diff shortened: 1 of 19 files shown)


src/components/labwareScanPanel/columns.tsx
```diff
@@ -22,6 +22,14 @@ const color: ColumnFactory<Map<number, any>> = (meta) => {
   };
 };
 
+function joinUnique(array: string[]) {
+  return Array.from(new Set<string>(array)).join(", ");
+}
+
+function valueFromSamples(labware: LabwareLayoutFragment, sampleFunction: (sample: any) => (string)) {
```
jbeast, line 29: Rather than using `any`, you could use TypeScript's [indexed access types](https://www.typescriptlang.org/docs/handbook/2/indexed-access-types.html) for the `sample` argument here. (`LabwareLayoutFragment["slots"][number]["samples"][number]` I think)

jbeast (approved):
Looks really good. The new `labwareCheckFunction` param is going to come in handy in the future (and probably should be added for some of the pages already built).

PR #49, opened 2021-07-21
add cassette labware type as an option in block registration

khelwood:


khelwood:
Tests are failing but that seems to be unrelated to the pull request.

PR #51, opened 2021-07-27
x558 Added History page

jbeast:
Add an export link to History component

Add mock API and tests for History

### DIFF
--- a/src/components/history/History.tsx
+++ b/src/components/history/History.tsx
@@ -131,10 +131,10 @@ const historyColumns: Array<Column<HistoryTableEntry>> = [
     accessor: "biologicalState",
   },
   {
-    Header: "Process State",
-    accessor: "processState",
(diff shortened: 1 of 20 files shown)


src/components/history/History.tsx
```diff
@@ -0,0 +1,157 @@
...
+  {
+    Header: "Section Number",
+    accessor: "sectionNumber",
+  },
+  {
+    Header: "Biological State",
+    accessor: "biologicalState",
+  },
+  {
+    Header: "Process State",
```
khelwood, line 134: Can we call it "labware state" instead of "process state"? Otherwise "process state" will infect the vocabulary of the software, and I don't think it's very meaningful.

jbeast: Yep, good idea. Have made the change to labware state.

khelwood (approved):

PR #64, opened 2021-09-08
x681 - Imaging

jbeast:


src/components/Routes.tsx
```diff
@@ -95,6 +96,24 @@ export function Routes() {
...
+        render={(routeProps) => (
+          <DataFetcher
+            key={routeProps.location.key}
+            dataFetcher={stanCore.GetRecordInPlaceInfo}
+          >
+            {(recordInPlaceInfo) => (
+              <RecordInPlace
+                title={"Imaging"}
+                operationType={"image"}
+                equipment={recordInPlaceInfo.equipments}
```
khelwood, line 110: I think you should only load equipment with the category "scanner" for the imaging page. That's what category was intended for.

jbeast: Thanks, good spot. Have changed it now.

khelwood (approved):

PR #70, opened 2021-10-05
X598

seenanair:
Closes #

x598 implementation and updates.
Notes :- 
- genericSearchMachine.ts is an exact copy of SearchMachine.ts (only change is SearchContext object fields) and one of these files can be later deleted after review. This is duplicated currently to avoid changing existing functionalities.

*
*
* ...


src/components/workProgress/WorkProgressInput.tsx
```diff
@@ -0,0 +1,236 @@
...
+            )}
+            <div className="space-y-2 md:px-10 md:space-y-0 md:flex md:flex-row md:justify-center md:items-center md:gap-4">
+              <div className="md:flex-grow">
+                <FormikSelect
+                  label=""
+                  name="type"
+                  data-testid={"type"}
+                  value={selectedType}
+                  onChange={(e: React.ChangeEvent<HTMLSelectElement>) => {
+                    setFieldValue(
```
jbeast, line 177: I'm not quite sure the purpose of the `setFieldValue` here. Isn't `"searchType"` always going to be `undefined`?

src/lib/machines/workProgress/genericSearchMachine.ts
```diff
@@ -0,0 +1,109 @@
...
+import { ClientError } from "graphql-request";
+import { Maybe } from "../../../types/sdk";
+import { GenericSearchService } from "../../services/genericSearchService";
+/**
+ * Context for Search Machine
+ */
+export interface SearchContext {
+  findRequest: any;
+  maxRecords?: number;
+  searchResult?: any;
```
jbeast, line 13: Instead of using `any` here, could it use a generic to keep type safety?

src/lib/machines/workProgress/genericSearchMachine.ts
```diff
@@ -0,0 +1,109 @@
...
+  serverError?: Maybe<ClientError>;
+}
+
+type FindEvent = { type: "FIND"; request: Object };
+type SearchDoneEvent = {
+  type: "done.invoke.search";
+  data: Object;
```
jbeast, line 20: This can also probably be a generic argument.

src/lib/services/genericSearchService.ts
```diff
@@ -0,0 +1,4 @@
+import { GenericSearchResultsType } from "../../types/stan";
+export interface GenericSearchService {
+  search(findRequest: any): Promise<GenericSearchResultsType>;
```
jbeast, line 3: I think this could also be a generic argument.

src/pages/WorkProgress.tsx
```diff
@@ -0,0 +1,234 @@
...
+  rowBDate: Date | undefined
+) => {
+  if (rowADate && rowBDate) {
+    return rowADate.getTime() - rowBDate.getTime();
+  }
+  if (rowADate) return 1;
+  return 0;
+};
+/**
+ * Sort functionality for Status. The status need to be sorted in the order "active", "completed", "paused", "failed"
```
jbeast, line 151: Was this one of the acceptance criteria? It may be overkill.

src/lib/machines/workProgress/genericSearchMachine.ts
seenanair: Good point. I even thought about it while implementing. But the searchService and workProgressService are using completely different structures. So I was not sure how to keep a generic structure in place. It would be great to have a chat about this sometime later.

src/pages/WorkProgress.tsx
seenanair: Yes, This has came up later during the standup by Minal. The rows has to be ordered by status in the order -active,complete,paused,failed.) and within status SGP/R&D numbers need to be sorted in descending order. Please suggest if you have any better ways to improve performance.

PR #73, opened 2021-10-08
X783

seenanair:
Closes # 783

1) Add columns to the SGP management page: 
*Number of blocks 
*Number of slides 
*editable number up 200. 
*Need to be able to be editable throughout as more blocks/slides might be added

2) Add Number of blocks and Number of slides fields while allocating a new SGP Number.

3) Add a status of ""Not started"" - default setting for once a work number is created

### DIFF
--- a/cypress/integration/pages/workProgressSpec.ts
+++ b/cypress/integration/pages/workProgressSpec.ts
@@ -23,8 +23,8 @@ describe("Work Progress", () => {
         before(() => {
           cy.findByTestId("type").select("Status");
         });
-        it("shows a drop down box with value active ", () => {
-          cy.get("[data-testid=valueSelect]").should("have.value", "active");
+        it("shows a drop down box with value unstarted ", () => {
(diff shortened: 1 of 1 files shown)


src/components/workAllocation/WorkRow.tsx
```diff
@@ -84,12 +88,64 @@ export default function WorkRow({
...
+        data-testid={workNumber + "-" + workNumValueType}
+        className={"border-0 border-gray-100 "}
+        type="number"
+        min="0"
+        max={MAX_NUM_BLOCKANDSLIDES}
+        step="1"
+        onChange={(e) => {
+          handleWorkNumValueChange(e.currentTarget.value, workNumValueType);
+        }}
+        value={workNumValue ?? ""}
```
jbeast, line 124: I think React will give a warning about controlled/uncontrolled inputs about this. May need to use `defaultValue` instead.

PR #80, opened 2021-10-29
X785- Splitting Labwork menu

seenanair:


jbeast:
Nice! About time the menu and menu items became their own components 😄

seenanair:
> Nice! About time the menu and menu items became their own components 😄



> Nice! About time the menu and menu items became their own components 😄

Thank you, Chris.

PR #83, opened 2021-11-04
X666 Visium Perm

jbeast:


khelwood (commented):
It looks like you are not supporting negative controls, only positive controls

jbeast:
Positive and negative controls are supported. It does however default to positive when a slot is selected as a control.

khelwood:
You're right! I couldn't tell that from the code, but I tried it out and it works. My mistake.

khelwood (approved):

PR #87, opened 2021-11-10
X631 update 2

seenanair:
- Analysis type change - RIN,DV200 -> RIN analysis,DV200 analysis
- Empty field in Comment dropdown
- SGP Number dropdown for all samples

### DIFF
--- a/src/components/extractResult/extractResult.machine.ts
+++ b/src/components/extractResult/extractResult.machine.ts
@@ -141,7 +141,6 @@ export const extractResultMachine = createMachine<
       }),
       assignServerError: assign((ctx, e) => {
         if (e.type !== "error.platform.extractResult") return;
-        debugger;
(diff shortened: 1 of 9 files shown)


src/components/extractResult/extractResult.machine.ts
```diff
@@ -141,6 +141,7 @@ export const extractResultMachine = createMachine<
       }),
       assignServerError: assign((ctx, e) => {
         if (e.type !== "error.platform.extractResult") return;
+        debugger;
```
jbeast, line 144: Oh no 😸

jbeast (changes requested):
Think it's just that one `debugger` statement. Looks good apart from that.

src/components/extractResult/extractResult.machine.ts
seenanair: Done.

PR #94, opened 2021-12-03
X697

jbeast:
New stain types.

### DIFF
--- a/src/pages/staining/ComplexStainForm.tsx
+++ b/src/pages/staining/ComplexStainForm.tsx
@@ -1,0 +1,255 @@
+import React from "react";
+import {
+  ComplexStainRequest,
+  RecordComplexStainMutation,
+  StainPanel,
+} from "../../types/sdk";
+import { Form, Formik } from "formik";
+import FormikInput from "../../components/forms/Input";
+import GrayBox, { Sidebar } from "../../components/layouts/GrayBox";
+import { motion } from "framer-motion";
+import variants from "../../lib/motionVariants";
+import Heading from "../../components/Heading";
+import MutedText from "../../components/MutedText";
+import LabwareScanPanel from "../../components/labwareScanPanel/LabwareScanPanel";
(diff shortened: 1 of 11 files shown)


src/pages/staining/ComplexStainForm.tsx
```diff
@@ -0,0 +1,255 @@
...
+    panel: Yup.string()
+      .oneOf(Object.values(StainPanel))
+      .required()
+      .label("Experimental Panel"),
+    labware: Yup.array()
+      .of(
+        Yup.object().shape({
+          barcode: Yup.string().required().label("Barcode"),
+          bondBarcode: Yup.string().required().label("Bond Barcode"),
+          bondRun: Yup.number().integer().positive().label("Bond Run"),
```
seenanair, line 85: In acceptance test it says -bond run number - 4 digit integer (+ve) . Is it required  to put 4 digit number constraint here?

seenanair (commented):
Also, in Acceptance test it says- Page will allow you to submit a request, then change the stain type (without losing the other information entered)

src/pages/staining/ComplexStainForm.tsx
```diff
@@ -0,0 +1,255 @@
...
+  const validationSchema = Yup.object().shape({
+    stainType: Yup.string().required().label("Stain Type"),
+    plex: Yup.number()
+      .integer()
+      .min(plexMin)
+      .max(plexMax)
+      .required()
+      .label("Plex Number"),
+    panel: Yup.string()
+      .oneOf(Object.values(StainPanel))
```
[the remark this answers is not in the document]
jbeast: Good point, well spotted. I'll change.

src/pages/staining/ComplexStainForm.tsx
jbeast: I thought that might be a bit too restrictive for the client. Seemed like the sort of thing that may need to change in the relatively near future, so thought I'd leave that as core side validation only.

PR #102, opened 2022-01-10
X821 - Priority field in SGP management page

seenanair:


khelwood:
The client currently only sends an update work priority request if the work status is "Unstarted". In any other status, editing the work priority doesn't send a request.
It should work in any work status except for "completed" and "failed". For those, the UI should not let the user try and update the priority.

seenanair:
> The client currently only sends an update work priority request if the work status is "Unstarted". In any other status, editing the work priority doesn't send a request. It should work in any work status except for "completed" and "failed". For those, the UI should not let the user try and update the priority.

Ok. I didn't know that such a condition existed. I will fix that.

seenanair:
> The client currently only sends an update work priority request if the work status is "Unstarted". In any other status, editing the work priority doesn't send a request. It should work in any work status except for "completed" and "failed". For those, the UI should not let the user try and update the priority.

Fixed and pushed changes.

khelwood:
When a work status is changed to "completed" or "failed", the core side clears its priority; but the client does not show that until you reload the page.

seenanair:
> When a work status is changed to "completed" or "failed", the core side clears its priority; but the client does not show that until you reload the page.

Fixed

PR #109, opened 2022-01-24
Change in Replicate number format

seenanair:
Replicate number  is now a string comprising 
-An integer (>0) and 
-A single lower case letter (optional)

### DIFF
--- a/src/lib/factories/sampleFactory.ts
+++ b/src/lib/factories/sampleFactory.ts
@@ -27,11 +27,11 @@ export const sampleFactory = Factory.define<Sample>(
 );
 
 export const tissueFactory: Factory<Tissue> = Factory.define<Tissue>(
-  ({ sequence, params, associations }) => ({
+  ({ params, associations }) => ({
     __typename: "Tissue",
     externalName:
       params.externalName ?? `${faker.name.lastName()}${faker.random.number()}`,
-    replicate: params.replicate ?? _.random(10),
+    replicate: params.replicate ?? String(_.random(10)),
     spatialLocation:
       associations.spatialLocation ?? spatialLocationFactory.build(),
     donor: associations.donor ?? donorFactory.build(),
@@ -72,7 +72,7 @@ export const mouldSizeFactory: Factory<MouldSize> = Factory.define<MouldSize>(
 );
 
 export const hmdmcFactory: Factory<Hmdmc> = Factory.define<Hmdmc>(
-  ({ sequence, params }) => ({
+  ({ params }) => ({
     __typename: "Hmdmc",
     hmdmc: params.hmdmc ?? `${_.random(1, 99)}-${_.random(100, 10000)}`,
     enabled: params.enabled ?? true,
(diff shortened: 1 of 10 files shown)


src/lib/factories/sampleFactory.ts
```diff
@@ -27,11 +27,11 @@ export const sampleFactory = Factory.define<Sample>(
 );
 
 export const tissueFactory: Factory<Tissue> = Factory.define<Tissue>(
-  ({ sequence, params, associations }) => ({
```
khelwood, line 30: What does it mean that these `sequence` parameters have been removed?

src/pages/Registration.tsx
```diff
@@ -45,7 +45,7 @@ export function getRegistrationFormBlock(): RegistrationFormBlock {
     clientId: Date.now(),
     externalIdentifier: "",
     spatialLocation: -1, // Initialise it as invalid so user has to select something
-    replicateNumber: 0,
+    replicateNumber: "1",
```
khelwood, line 48: Can we have it default to "" please?

src/pages/SlideRegistration.tsx
```diff
@@ -113,7 +113,7 @@ function buildSample(): SlideRegistrationFormSection {
     tissueType: "",
     externalIdentifier: "",
     spatialLocation: 0,
-    replicateNumber: 0,
+    replicateNumber: "1",
```
khelwood, line 116: Can we have it default to "" please?

src/lib/factories/sampleFactory.ts
seenanair: @khelwood sequence parameter is not used in sampleFactory and in other places , so there were ts compile warnings about "Unused parameter sequence"

src/lib/validation/registrationValidation.ts
```diff
@@ -93,9 +93,11 @@ export default class RegistrationValidation {
   }
 
   get replicateNumber() {
-    return validation.requiredNumber({
+    return validation.requiredString({
       label: "Replicate Number",
-      min: 1,
+      restrictChars: /^[1-9]\d*[a-z]?$/,
+      errorMessage:
+        "Replicate Number contains invalid characters. Only numbers or a number followed by a lower case letter is permitted.",
```
khelwood, line 100: Can this say "Replicate Number must be a positive integer, optionally followed by a lower case letter." please

src/pages/Registration.tsx
seenanair: Done. Good point -That will actually remind the user to enter a valid value.

src/pages/SlideRegistration.tsx
seenanair: Done

src/lib/validation/registrationValidation.ts
seenanair: Done. Much better warning message than before.

khelwood (approved):

PR #110, opened 2022-01-28
RT#742040-Search page error

seenanair:
Key parameter was not provided to the Search page in Router which in turn caused the situation that xState machine is not fetching the right API

### DIFF
--- a/cypress/integration/pages/searchSpec.ts
+++ b/cypress/integration/pages/searchSpec.ts
@@ -63,42 +63,15 @@ describe("Search", () => {
       }
     );
 
-    context("when a search has more results than the requested amount", () => {
-      before(() => {
-        cy.visit("/search");
-
-        cy.msw().then(({ worker, graphql }) => {
-          worker.use(
-            graphql.query<FindQuery, FindQueryVariables>(
-              "Find",
-              (req, res, ctx) => {
-                return res.once(ctx.data({ find: buildFindResult(200, 150) }));
-              }
-            )
-          );
-        });
-
-        cy.findByLabelText("Donor ID").type("DNR123");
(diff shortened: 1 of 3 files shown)


cypress/integration/pages/searchSpec.ts
```diff
@@ -63,42 +63,15 @@ describe("Search", () => {
       }
     );
 
-    context("when a search has more results than the requested amount", () => {
```
khelwood, line 66: Why is this test no longer required?

khelwood (commented):
Seems to have fixed the problem. Can you explain the fix? My guess is some kind of stale data between the home and search pages.

seenanair:
> Seems to have fixed the problem. Can you explain the fix? My guess is some kind of stale data between the home and search pages.

A key parameter is required for react-router for each the paths navigated. The Search page (as well as Home page) was missing that which was causing those issues. All the data which needs to pre-fetched before routing to a page was using a common state machine (xState) using a component DataFetcher. The default home page is fetching data using workTypes API and when we switch to Search page, since both of them don't have keys( which act as an identifier) the xState machine get confused and instead of calling the tissueTypes API , workTypes data is passed on Search page. The same was happening vice versa while visiting Home page from a Search page

cypress/integration/pages/searchSpec.ts
seenanair: Good check. This test is required.
 With the changes in Search page, the test cases using mock service data were failing  which took me a while to figure out and in the process of trying out different options, I think I somehow deleted it.  With Mock server 'once' API call, the mocked data was not returning which I didn't understand why. ('once' API  is used to mark the current request handler as used and to ignore any subsequent request matches to this handler. )

PR #111, opened 2022-02-02
Aliquot a tube of solution out into multiple new tubes

seenanair:
Scan tube barcode in
Select number of new labels needed
Discard source barcode as sample is transferred to new tubes
Operation is recorded by the user in one step


src/pages/Aliquot.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+          <div className="my-4 mx-4 sm:mx-auto p-4 rounded-md bg-gray-100">
+            <p className="my-3 text-gray-800 text-sm leading-normal">
+              Once <span className="font-bold text-gray-900">source tube</span>{" "}
+              is scanned and
+              <span className="font-bold text-gray-900">
+                {" "}
+                number of destination tubes{" "}
+              </span>{" "}
+              are entered, <p>click Aliquot to create destination labwares.</p>
```
khelwood, line 240: "labwares" isn't a word. "tubes" would be fine here, since we know what specific type the destinations are.

khelwood (approved):

PR #112, opened 2022-02-07
X832 - Download option in Dashboard page

seenanair:
- Downloads data in sorted order
- Fixed the downloaded file name issue (Please test with different browsers)

### DIFF
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -5,6 +5,7 @@ import { GridDirection, Maybe } from "../types/sdk";
 import { HasEnabled, SizeInput } from "../types/stan";
 import _ from "lodash";
 import { Key } from "react";
+import { Column } from "react-table";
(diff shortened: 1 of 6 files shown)


src/lib/helpers.ts
```diff
@@ -246,3 +247,54 @@ export function getEnumKeyByEnumValue<T extends { [index: string]: string }>(
...
+          .map((column) => {
+            if (typeof column.accessor === "string") {
+              const value = entry[column.accessor];
+              if (
+                typeof value === "object" &&
+                (value as Object) instanceof Date
+              ) {
+                const date = value as Date;
+                return date.toLocaleDateString();
+              } else if (value === "number") {
```
khelwood, line 284: Shouldn't this be checking typeof value ?

khelwood (approved):

PR #124, opened 2022-03-02
X678_stainqc_tissuecoverage - Record "Tissue coverage" on the Stain QC page

seenanair:
"Tissue coverage" on the Stain QC pageTissue coverage is optional.

- Tissue coverage is optional
-  For all occupied slots on the slide the user can assign an integer  value for Tissue Coverage  (range 0 to 100 inclusive)
- Tissue coverage unit is %

### DIFF
--- a/src/components/labwareResult/LabwareResult.tsx
+++ b/src/components/labwareResult/LabwareResult.tsx
@@ -72,7 +72,8 @@ export default function LabwareResult({
 
   /**Ensure Tissue Coverage value is in the range 0 to 100 inclusive**/
   const validateMeasurementField = (value: string) => {
-    const regEx = /^(^$|[0-9]|([1-9][0-9])|100)$/;
+    if (value.length === 0) return true;
+    const regEx = /^[0-9]+$/;
     const val = regEx.test(value);
     if (val) {
       const coverage = Number(value);
--- a/src/components/labwareResult/LabwareResult.tsx
+++ b/src/components/labwareResult/LabwareResult.tsx
@@ -1,7 +1,6 @@
 import React, { useEffect } from "react";
 import Labware from "../labware/Labware";
 import { Select } from "../forms/Select";
-import { optionValues } from "../forms";
 import BlueButton from "../buttons/BlueButton";
 import PinkButton from "../buttons/PinkButton";
 import createLabwareResultMachine from "./labwareResult.machine";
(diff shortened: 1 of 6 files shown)


src/components/labwareResult/LabwareResult.tsx
```diff
@@ -56,64 +58,109 @@ export default function LabwareResult({
...
 
   useEffect(() => {
     onChange(labwareResult);
   }, [labwareResult, onChange]);
 
+  const isMeasurementExist = "slotMeasurements" in labwareResult;
+
+  /**Ensure Tissue Coverage value is in the range 0 to 100 inclusive**/
+  const validateMeasurementField = (value: string) => {
+    const regEx = /^(^$|[0-9]|([1-9][0-9])|100)$/;
```
khelwood, line 75: You're about to try and convert the value to a number, so accepting the empty string `^$` does not make sense here. `/^([0-9]|([1-9][0-9])|100)$/` would be OK, but you could simply use `/^[0-9]+$/` to check it's made up of one or more digits.

src/components/labwareResult/LabwareResult.tsx
```diff
@@ -56,64 +58,109 @@ export default function LabwareResult({
...
-                }`}
-                onClick={() => {
-                  send({ type: "PASS", address: slot.address });
-                }}
-              />
+      isSlotFilled(slot) && (
+        <div
+          className={`flex flex-col ${
+            isMeasurementExist &&
+            labware.slots.length > 1 &&
```
khelwood, line 90: Why must labware.slots.length be greater than 1 here?

seenanair: That is for the UI style. If there are more than one slot there will be separator line displayed between each slot to make it more clear which is not needed if there is only one slot.

src/components/labwareResult/LabwareResult.tsx
seenanair: Thought this would avoid entering values like 05 etc  to give visual effect. But as you said functionality wise that won't be a problem since we are converting to an integer. 
I have put empty string option because acceptance test says it can be an empty field. So if the user enters a value and then deletes it or use backspace it will allow to remove the text completely. 
But again, for this case, we can always check the string length to 0 and return a true without using regex.

seenanair: @khelwood Updated regex and pushed in changes.

khelwood (approved):

PR #129, opened 2022-03-14
X846 auto section number

seenanair:
Auto numbering option for section numbers
- Option to switch between Auto and Manual
- Default option is 'Auto'
- In 'Auto mode, 
        - All section numbers will be numbered starting from highest section number for the source
        - All section number fields will be disabled
        - Adding/Deleting new sections updates the section numbers automatically
        - If a Tube is cancelled, section numbers will be removed from the cancelled ones and all other labware for that source will be re-numbered
        - If any labware other than Tube is deleted,  a warning will be given about renumbering of section numbers

### DIFF
--- a/src/components/sectioningConfirm/ConfirmLabware.tsx
+++ b/src/components/sectioningConfirm/ConfirmLabware.tsx
@@ -22,14 +22,21 @@ import {
 } from "../../pages/sectioning";
 import { createConfirmLabwareMachine } from "./confirmLabware.machine";
 import { LayoutPlan } from "../../lib/machines/layout/layoutContext";
-import { CommentFieldsFragment, ConfirmSectionLabware } from "../../types/sdk";
+import {
+  CommentFieldsFragment,
+  ConfirmSectionLabware,
+  LabwareFieldsFragment,
+} from "../../types/sdk";
 import { selectConfirmOperationLabware } from "./index";
 import RemoveButton from "../buttons/RemoveButton";
 
 interface ConfirmLabwareProps {
   originalLayoutPlan: LayoutPlan;
   comments: Array<CommentFieldsFragment>;
-  onChange: (labware: ConfirmSectionLabware) => void;
+  onChange: (
+    labware: ConfirmSectionLabware,
+    sourceLabwares: LabwareFieldsFragment[]
+  ) => void;
   onRemoveClick: (labwareBarcode: string) => void;
   disableSectionNumbers?: boolean;
 }
@@ -66,9 +73,14 @@ const ConfirmLabware: React.FC<ConfirmLabwareProps> = ({
 
   useEffect(() => {
     if (confirmOperationLabware) {
-      onChange(confirmOperationLabware);
+      onChange(
+        confirmOperationLabware,
+        Array.from(layoutPlan.plannedActions.values()).flatMap((sources) =>
+          Array.from(sources.values()).map((source) => source.labware)
+        )
+      );
     }
-  }, [onChange, confirmOperationLabware]);
+  }, [onChange, confirmOperationLabware, layoutPlan.plannedActions]);
 
   /***Update section numbers whenever there is an update in section numbers in parent**/
   useEffect(() => {
@@ -83,7 +95,10 @@ const ConfirmLabware: React.FC<ConfirmLabwareProps> = ({
       className="relative p-3 shadow"
     >
       <RemoveButton onClick={() => onRemoveClick(labware.barcode!)} />
-      <div className="md:grid md:grid-cols-2">
+      <div
+        data-testid={`div-slide-${labware.barcode}`}
+        className="md:grid md:grid-cols-2"
+      >
         <div className="py-4 flex flex-col items-center justify-between space-y-8">
           <Labware
             labware={labware}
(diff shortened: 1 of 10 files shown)


src/components/sectioningConfirm/SectioningConfirm.tsx
```diff
@@ -121,18 +163,47 @@ export default function SectioningConfirm({
...
-                    <DataTable
-                      data={sourceLabware}
-                      columns={[
-                        columns.barcode(),
-                        columns.highestSectionForSlot("A1"),
-                      ]}
-                    />
-                  </div>
-                )}
+                  <>
```
khelwood, line 166: What does the addition of <> here signify?

src/components/sectioningConfirm/SectioningConfirm.tsx
```diff
@@ -189,6 +273,35 @@ export default function SectioningConfirm({
...
+                            labwarePlanToDelete.destinationLabware.barcode
+                          );
+                        setLabwarePlanToDelete(undefined);
+                      },
+                    },
+                  ]}
+                >
+                  <p className={"font-bold mt-8"}>
+                    Section numbers of remaining plans with same source labware
+                    will be renumbered with deletion.
```
khelwood, line 302: This is not very clear. I'm not sure what being "renumbered with deletion" means.

src/components/sectioningConfirm/confirmLabware.machine.ts
```diff
@@ -159,6 +177,12 @@ export const createConfirmLabwareMachine = (
             UPDATE_SECTION_NUMBER: {
               actions: ["updateSectionNumber", "commitConfirmation"],
             },
+            UPDATE_ALL_SECTION_NUMBERS: {
+              actions: ["updateAllSectionNumbers", "commitConfirmation"],
+            },
+            SET_MIN_SECTION_NUMBER: {
+              actions: ["setMinimumSectionNumber", "commmitConfirmation"],
```
khelwood, line 184: "commmitConfirmation" should presumably be "commitConfirmation"

khelwood (commented):
I have made a couple of observations here about the source code. I will message on slack to talk about some other things I found when testing.

src/components/sectioningConfirm/SectioningConfirm.tsx
seenanair: In React, expressions must have one parent element. So as to wrap that entire segment within a parent <> </> is used, or we can use <React.Fragment> </React.Fragment> as a wrapper.

src/components/sectioningConfirm/SectioningConfirm.tsx
seenanair: When a slide is deleted, all the section numbers in other labwares, which is sectioned from the same source, will be updated. Please suggest, if any better wordings.

src/components/sectioningConfirm/confirmLabware.machine.ts
seenanair: Typo- will correct.

src/components/sectioningConfirm/SectioningConfirm.tsx
khelwood: I see. I think "planned section numbers of other labware will be updated" would cover it. Why does the warning only apply to slides, not to tubes?

seenanair: Tubes cannot be deleted but can only be cancelled. Minal mentioned to give a warning only while deleting Slides. But again, please check and suggest whether it is needed to give a warning message for 'cancelling' tubes as well.

khelwood: I'm not sure what distinction the users draw between something being "deleted" and "cancelled": I think the behaviour should be consistent.

seenanair: Ok, I will put that in. While cancelling, the labware is not removed from UI and this operation is revertible.

src/components/sectioningConfirm/ConfirmLabware.tsx
```diff
@@ -25,19 +25,31 @@ import { LayoutPlan } from "../../lib/machines/layout/layoutContext";
...
-  onRemoveClick: (labwareBarcode: string) => void;
+  onSectionUpdate: (layoutPlan: LayoutPlan) => void;
+  onSectionNumberChange: (
+    layoutPlan: LayoutPlan,
+    slotAddress: string,
+    sectionIndex: number,
+    sectionNumber: number
+  ) => void;
+  removePlan: (barcode: string) => void;
+  autoMode: boolean;
```
khelwood, line 42: How come the mode here is represented as a boolean instead of your SectionNumberMode enum?

src/components/sectioningConfirm/ConfirmLabware.tsx
```diff
@@ -108,7 +159,8 @@ const ConfirmLabware: React.FC<ConfirmLabwareProps> = ({
                   key={slot.address}
                   slot={slot}
                   value={addressToCommentMap.get(slot.address) ?? ""}
-                  disabled={current.matches("done")}
+                  disabledComment={current.matches("done")}
+                  disabledSectionNumber={autoMode}
```
khelwood, line 163: Previously the section number would be disabled if `current.matches("done")`, but now it will only be disabled in automode. Should this be `disabledSectionNumber={autoMode || current.matches("done")}` ?

seenanair: Yes, you are right.

src/components/sectioningConfirm/ConfirmLabware.tsx
seenanair: Updated to enum

khelwood (approved):

PR #131, opened 2022-03-24
X838 dual index plate

seenanair:
Dual index plate" feature in the Visium Menu to record the transfer of dual-index reagent to our 96 well plates containing samples

1. User must select work number
2. User scans in a dual-index plate barcode which is a 24 digit number
3. User scans in a 96 well plate
4. The filled slots of 96 well plate should be displayed in green
5. The user will transfer a number of slots from the dual index plate to 96 well plate. Each source slot can only be used once.
6. The first time the user tries to use a dual index plate, the system creates it.
7. Subsequent uses of dual index plate indicate which slots are empty (greys them out in the UI) based on previous uses, otherwise in blue.
8. Transferring to an empty slot in 96 well plate will not be allowed and a warning will be given.
9. All transferred slot information is displayed in a 'Mapping' table

### DIFF
--- a/cypress/integration/pages/dualIndexPlateSpec.ts
+++ b/cypress/integration/pages/dualIndexPlateSpec.ts
@@ -187,7 +187,15 @@ describe("Dual Index Plate", () => {
       });
       it("should display a success message", () => {
         cy.findByText("Reagents transferred").should("be.visible");
+      });
+      it("should display Reset button", () => {
         cy.findByRole("button", { name: /Reset Form/i }).should("be.visible");
+      });
+      it("should disable Clear button", () => {
+        cy.findByRole("button", { name: /Clear/i }).should("be.disabled");
+      });
+      it("should disable Remove button", () => {
+        cy.findAllByTestId("removeButton").eq(0).should("be.disabled");
       });
     });
   });
--- a/cypress/integration/pages/dualIndexPlateSpec.ts
+++ b/cypress/integration/pages/dualIndexPlateSpec.ts
@@ -91,14 +91,42 @@ describe("Dual Index Plate", () => {
         );
       });
     });
-    context("when user clears the mapped slot", () => {
+  });
+  describe("On removing mapping", () => {
+    before(() => {
+      cy.get("#sourceLabwares").within(() => {
+        cy.findByText("A3").click();
+      });
+      cy.get("#destLabwares").within(() => {
+        cy.findByText("A3").click();
+      });
+    });
+    it("should display two mappings in table", () => {
+      //find("tr) return rows including header
+      cy.findByRole("table").find("tr").should("have.length", 3);
+    });
+    context("when user clicks remove button in the mapping table", () => {
       before(() => {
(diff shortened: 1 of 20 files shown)


khelwood:
Screenshot 2022-03-28 at 12 05 22

seenanair:
@khelwood The grey rectangle is for displaying the 'Clear button'. I have used the same style from Visium cDNA page  in Dual Index Plate, where both have some is similarities in transferring between slots in labware. Visium cDNA has a 'Clear' button in that which is displayed in a grey bar, so, I used the same style. Do you think it need to be changed?
I will change the mapping table insets, but I didn't intend to move that to grey bar. Do you think it is not very clear?

seenanair:
> It looked to me like something was supposed to be in that rectangle but was not being shown. But if the users see it and don't have a problem with it, then it is fine.

I have a look into that and I think you are right. I will move the table below that labware display above    'Clear' button. Hope that make sense.

seenanair:
> It looked to me like something was supposed to be in that rectangle but was not being shown. But if the users see it and don't have a problem with it, then it is fine.

Also, you didn't mention about grey color bar. Do you have any other options to suggest?

khelwood:
If you mean this grey bar

khelwood:
If you mean this grey bar
Screenshot 2022-03-28 at 12 05 22
then I'm not sure why it is there or what it signifies

seenanair:
Ohh! I didn't notice that. that is not displayed in mine. Are using Chrome?
Screenshot 2022-03-28 at 15 32 09
?

khelwood:
I am using Firefox

seenanair:
Checked in Firefox. But that bar is not getting displayed

Screenshot 2022-03-28 at 15 39 48

khelwood:
I tried again and the grey bar isn't showing up, so I guess it was a glitch.
I found that "Clear" only removed one mapping from the table (the mapping to the selected slot?). I had thought "Clear" would clear the whole table. Can I suggest that the clearer way to delete one mapping from the table should be an X button next to that row in the table?

seenanair:
> I tried again and the grey bar isn't showing up, so I guess it was a glitch. I found that "Clear" only removed one mapping from the table (the mapping to the selected slot?). I had thought "Clear" would clear the whole table. Can I suggest that the clearer way to delete one mapping from the table should be an X button next to that row in the table?

Yes, Clear only removes the selected one(mapping). Again, it was how the clear works in Visium cDNA. So I thought it would be ok. But, your suggestion seems to be good for a better UI interaction. So, I will implement that.

seenanair:
> I tried again and the grey bar isn't showing up, so I guess it was a glitch. I found that "Clear" only removed one mapping from the table (the mapping to the selected slot?). I had thought "Clear" would clear the whole table. Can I suggest that the clearer way to delete one mapping from the table should be an X button next to that row in the table?

@khelwood I have pushed in these changes. Please test and review the changes when you get time

khelwood (approved):

PR #144, opened 2022-05-03
x863 Add "Desiccator" to store page

khelwood:
Depends on storelight patch x863_new_locations.sql

### DIFF
--- a/src/static/store.json
+++ b/src/static/store.json
@@ -8,7 +8,7 @@
         "barcode": "STO-003D"
       },
       {
-        "barcode": "STO-12DC"
+        "barcode": "STO-12CD"
       }
     ],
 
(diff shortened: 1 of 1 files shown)


src/static/store.json
```diff
@@ -6,6 +6,9 @@
       },
       {
         "barcode": "STO-003D"
+      },
+      {
+        "barcode": "STO-12DC"
```
seenanair, line 11: Is it STO-12CD instead of STO-12DC?
In sql , it is as follows
"_INSERT INTO location
  (barcode, name)
VALUES
  (**'STO-12CD', 'Dessicator'**_"  And 'Rack 1' is STO-12DC

khelwood: Well spotted! Thanks

seenanair (approved):

PR #149, opened 2022-06-09
X886 pot processing

seenanair:
Pot Processing feature  to allow for the division of an original sample into different types of pots of fixative or fetal waste labware, which will have no replicate number
(Fixative field is "None" for Fetal waste labware)


src/components/originalSampleProcessing/potProcessing/PotProcessing.tsx
```diff
@@ -0,0 +1,424 @@
...
+        .required()
+        .min(1),
+    });
+  }
+
+  /**Reformat the form data as mutation input**/
+  const buildTissuePotRequest = (
+    formData: PotFormData
+  ): PotProcessingRequest => {
+    debugger;
```
khelwood, line 283: debugger

khelwood (approved):

PR #151, opened 2022-06-16
Changes associated with renaming of 'solutionSample' to 'solution'

seenanair:
All 'solutionSample' fields are renamed to 'solution'


src/pages/registration/RegistrationForm.tsx
```diff
@@ -278,15 +278,15 @@ const RegistrationForm = <T extends TissueValues<B>, B>({
                             )}
                           </FormikSelect>
                         )}
-                        {"solutionSample" in block && (
+                        {"solution" in block && (
                           <FormikSelect
                             emptyOption
                             label="Solution Sample"
```
khelwood, line 284: I think this should say "Solution"

src/lib/validation/registrationValidation.ts
```diff
@@ -33,10 +33,10 @@ export default class RegistrationValidation {
     });
   }
 
-  get solutionSample() {
+  get solution() {
     return validation.requiredString({
       label: "Solution Sample",
```
khelwood, line 38: I think the label here should be Solution, not Solution Sample

src/pages/registration/RegistrationForm.tsx
khelwood: Anywhere else too that says "solution sample"

seenanair: Sorry, I didn't get that.

seenanair: Changes done 👍

khelwood (approved):

PR #152, opened 2022-06-20
FFPE Processing

seenanair:
Records the type of processing cycle run on the sample


cypress/integration/shared/registration.ts
```diff
@@ -166,7 +165,7 @@ export function shouldBehaveLikeARegistrationForm(
         cy.findByText("Medium is a required field").should("be.visible");
       });
     } else {
-      it("requires Soultion Sample", () => {
+      it("requires Soultion", () => {
```
khelwood, line 168: spelling: "Solution"

seenanair: Updated 👍

PR #153, opened 2022-06-22
X874 solution transfer

seenanair:
New feature to record operation of Solution transfer of samples


cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
khelwood, line 1: Is this file actually needed?

seenanair: No, it's not needed. Deleted 👍

khelwood (approved):

PR #154, opened 2022-06-23
X760 printing error

seenanair:
Printing error fix for Block processing
- Remove pre-barcoded tubes from labware before printing

### DIFF
--- a/src/components/originalSampleProcessing/ProcessingSuccess.tsx
+++ b/src/components/originalSampleProcessing/ProcessingSuccess.tsx
@@ -11,6 +11,7 @@ import { Column } from "react-table";
 import Heading from "../Heading";
 import { ModalBody } from "../Modal";
 import BlueButton from "../buttons/BlueButton";
+import { LabwareTypeName } from "../../types/stan";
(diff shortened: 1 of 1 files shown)


src/components/originalSampleProcessing/ProcessingSuccess.tsx
```diff
@@ -34,7 +35,11 @@ const ProcessingSuccess: React.FC<ProcessingSuccessProps> = ({
           variants={variants.fadeInWithLift}
           className="sm:max-w-xl w-full border-gray-200 p-4 rounded-md bg-gray-100 shadow"
         >
-          <LabelPrinter labwares={labware} />
+          <LabelPrinter
+            labwares={labware.filter(
+              (lw) => lw.labwareType.name !== LabwareTypeName.PRE_BARCODED_TUBE
```
khelwood, line 40: Would it make more sense to check if the labware type has a _labelType_ instead of checking its name?

seenanair: Pushed in changes.

khelwood (approved):

PR #155, opened 2022-06-27
X900 dependency update

seenanair:
Dependency update changes
Tailwind update - No need of purge section
xState update - id field is mandatory for service requests
Cypress update - cy.ts file extensions
React update  - Namespace tags not allowed in React 18, so Logo file is rewritten with json attributes


src/lib/machines/locations/locationMachine.ts
```diff
@@ -137,6 +136,7 @@ export const machineOptions: Partial<MachineOptions<
       if (e.type !== "STORE_BARCODE") {
         return Promise.reject();
       }
+      debugger;
```
khelwood, line 139: debugger

khelwood (commented):
I don't understand all the changes but I don't see any problem, other than the debugger

khelwood (approved):

PR #162, opened 2022-07-14
Fixed bug where comments were getting added to the request

BenTopping:
Found this bug when trying to create data to test the original sample processing views with

### DIFF
--- a/src/components/originalSampleProcessing/blockProcessing/BlockProcessingLabwarePlan.tsx
+++ b/src/components/originalSampleProcessing/blockProcessing/BlockProcessingLabwarePlan.tsx
@@ -206,7 +206,7 @@ const BlockProcessingLabwarePlan = React.forwardRef<
                 </FormikSelect>
                 <FormikSelect
                   label={"Processing comments"}
-                  name={`plans.${rowIndex}.comments`}
+                  name={`plans.${rowIndex}.commentId`}
                   emptyOption={true}
                 >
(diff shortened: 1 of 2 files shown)


seenanair (commented):
Great. Good Finding 👍

khelwood:
Has someone tried this out locally and confirmed that it now works?

seenanair:
> Has someone tried this out locally and confirmed that it now works?

@khelwood I am going to merge this along with others and going to test all of it together. Is it ok?

khelwood:
> > Has someone tried this out locally and confirmed that it now works?
> 
> @khelwood I am going to merge this along with others and going to test all of it together. Is it ok?

Since it appears to be addressing a live bug, I am going to merge it into UAT instead of devel, so we can get it tested and released immediately.

seenanair:
> > > Has someone tried this out locally and confirmed that it now works?
> > 
> > 
> > @khelwood I am going to merge this along with others and going to test all of it together. Is it ok?
> 
> Since it appears to be addressing a live bug, I am going to merge it into UAT instead of devel, so we can get it tested and released immediately.

Ok, I will test it and let you know

khelwood:
I think Ben's already tested it, so I am going to merge into uat and deploy to uat.

seenanair:
> I think Ben's already tested it, so I am going to merge into uat and deploy to uat.

Ok  👍

PR #165, opened 2022-07-18
x864 - Spatial locations show the name for registration

BenTopping:
Updated spatial location dropdown values in registration pages to include code and name

### DIFF
--- a/src/pages/registration/RegistrationForm.tsx
+++ b/src/pages/registration/RegistrationForm.tsx
@@ -220,8 +220,8 @@ const RegistrationForm = <T extends TissueValues<B>, B>({
                           label="Spatial Location"
                           name={`tissues.${currentIndex}.blocks.${blockIndex}.spatialLocation`}
                         >
-                          {availableSpatialLocations.map((spatialLocation, index) => (
-                            <option value={spatialLocation.code} key={index}>
+                          {availableSpatialLocations.map((spatialLocation) => (
+                            <option value={spatialLocation.code} key={spatialLocation.code}>
                               {spatialLocation.code + " - " + spatialLocation.name}
                             </option>
                           ))}
--- a/src/pages/registration/RegistrationForm.tsx
+++ b/src/pages/registration/RegistrationForm.tsx
@@ -220,11 +220,11 @@ const RegistrationForm = <T extends TissueValues<B>, B>({
                           label="Spatial Location"
                           name={`tissues.${currentIndex}.blocks.${blockIndex}.spatialLocation`}
                         >
-                          {optionValues(
-                            availableSpatialLocations,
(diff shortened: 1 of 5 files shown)


src/pages/registration/RegistrationForm.tsx
```diff
@@ -220,11 +220,11 @@ const RegistrationForm = <T extends TissueValues<B>, B>({
                           label="Spatial Location"
                           name={`tissues.${currentIndex}.blocks.${blockIndex}.spatialLocation`}
                         >
-                          {optionValues(
-                            availableSpatialLocations,
-                            "code",
-                            "code"
-                          )}
+                          {availableSpatialLocations.map((spatialLocation, index) => (
+                            <option value={spatialLocation.code} key={index}>
```
seenanair, line 224: It might be better to use 'code' as key, as it is unique.
React documentation says 

_> Keys
> In order to solve this issue, React supports a key attribute. When children have keys, React uses the key to match children in the original tree with children in the subsequent tree. For example, adding a key to our inefficient example above can make the tree conversion efficient:
> 
> <ul>
>   <li key="2015">Duke</li>
>   <li key="2016">Villanova</li>
> </ul>
> 
> <ul>
>   <li key="2014">Connecticut</li>
>   <li key="2015">Duke</li>
>   <li key="2016">Villanova</li>
> </ul>
> Now React knows that the element with key '2014' is the new one, and the elements with the keys '2015' and '2016' have just moved.
> 
> In practice, finding a key is usually not hard. The element you are going to display may already have a unique ID, so the key can just come from your data:
> 
> <li key={item.id}>{item.name}</li>
> When that’s not the case, you can add a new ID property to your model or hash some parts of the content to generate a key. The key only has to be unique among its siblings, not globally unique.
> 
> As a last resort, you can pass an item’s index in the array as a key. This can work well if the items are never reordered, but reorders will be slow.
> 
> Reorders can also cause issues with component state when indexes are used as keys. Component instances are updated and reused based on their key. If the key is an index, moving an item changes it. As a result, component state for things like uncontrolled inputs can get mixed up and updated in unexpected ways.
> 
> Here is [an example of the issues that can be caused by using indexes as keys](https://reactjs.org/redirect-to-codepen/reconciliation/index-used-as-key) on CodePen, and here is [an updated version of the same example showing how not using indexes as keys will fix these reordering, sorting, and prepending issues]. [](https://reactjs.org/redirect-to-codepen/reconciliation/no-index-used-as-key)_

BenTopping: Yep makes sense, the index is ok but as the doc says - use it as a last resort 👍

seenanair (approved):

PR #166, opened 2022-07-18
X852 project name and owner

seenanair:
Display of project name and project owner to appear next to the SGP number

### DIFF
--- a/src/components/WorkNumberSelect.tsx
+++ b/src/components/WorkNumberSelect.tsx
@@ -114,6 +114,7 @@ export default function WorkNumberSelect({
         emptyOption={true}
         onChange={handleWorkNumberChange}
         className={"flex-grow w-full"}
+        data-testid={"workNumber"}
       >
         {optionValues(works, "workNumber", "workNumber")}
       </FormikSelect>
@@ -134,6 +135,7 @@ export default function WorkNumberSelect({
           onChange={handleWorkNumberChange}
           emptyOption={true}
           onBlur={validateWorkNumber}
+          data-testid={"workNumber"}
         >
           {optionValues(works, "workNumber", "workNumber")}
         </Select>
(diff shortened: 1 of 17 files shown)


src/components/WorkNumberSelect.tsx
```diff
@@ -52,60 +59,95 @@ export default function WorkNumberSelect({
...
-    <>
-      <Select
-        value={selectedWorkNumber}
-        onChange={handleWorkNumberChange}
+    <div className={"flex flex-col"}>
+      <FormikSelect
+        label={label ?? ""}
+        name={name}
         emptyOption={true}
-        onBlur={validateWorkNumber}
```
BenTopping, line 105: Is there a reason this was removed? This will flash up the 'SGP number is required' error message onBlur. Think its still needed for components that arent using yup / formik

src/lib/helpers.ts
```diff
@@ -93,7 +93,7 @@ export function safeParseQueryString<T>(
     try {
       const castValue = params.schema.cast(parsed);
       return params.schema.isValidSync(castValue)
-        ? (castValue as unknown as T) ?? null
+        ? ((castValue as unknown) as T) ?? null
         : null;
```
BenTopping, line 97: Were these automated changes or did you make these? Not entirely sure what they are used for but just checking they are suppose to be included

BenTopping (approved):
Looks good, just a few questions but otherwise happy to approve 👍

src/components/WorkNumberSelect.tsx
seenanair: Editing error, it seems. I have put that back 👍

src/lib/helpers.ts
seenanair: Reverted it back 👍

seenanair:
> Looks good, just a few questions but otherwise happy to approve 👍

@BenTopping Thank you. I have updated those

PR #172, opened 2022-07-26
X911 dep update

seenanair:
Yarn library version update
Replacement of faker library with faker-js as faker is deprecated 
Node version update


package.json
```diff
@@ -87,30 +87,29 @@
...
+    "cypress": "^10.3.1",
     "cypress-intellij-reporter": "^0.0.7",
-    "faker": "^5.5.3",
     "fishery": "^2.2.2",
-    "msw": "^0.42.2",
+    "msw": "^0.44.2",
     "npm-run-all": "^4.1.5",
-    "postcss-cli": "^9.1.0",
+    "postcss-cli": "^10.0.0",
     "prettier": "2.7.1",
```
BenTopping, line 113: Noticed we do have prettier as a dependency, is there a protocol for using prettier for linting the code? If so maybe need to add some instructions to readme.

BenTopping (approved):
Most of these changes look auto generated so I am going to assume they are safe. Pulled down and tried a few things locally and worked fine

seenanair:
> Most of these changes look auto generated so I am going to assume they are safe. Pulled down and tried a few things locally and worked fine

Yes. There are some changes needed with the update of Yup library and replacement of faker library with js-faker. Hope you have checked those as well.,

package.json
seenanair: At present there is not one I think  and  no formatting error check is done using git hooks as well.  But, it is added as a dependency module in IDE which will enforce prettier formatting automatically.
Do you think, it is good to give this prettier check as a pre-commit hook ?

seenanair: Ok, I will try to add that as  a pre-commit hook as part of this story

README.md
```diff
@@ -7,8 +7,8 @@ This project was bootstrapped with [Create React App](https://github.com/faceboo
 
 ## Dependencies
 
-- `Node v14.19.3`
-- `Yarn v2.10.0`
+- `Node v16.15.1`
+- `Yarn v1.22.11`
```
BenTopping, line 12: were you planning to add something about git hooks / linting to readme ?

package.json
```diff
@@ -87,32 +89,33 @@
...
-    "start-server-and-test": "^1.11.5"
+    "start-server-and-test": "^1.11.5",
+    "husky": "^8.0.0",
+    "precise-commits": "^1.0.2"
```
BenTopping, line 118: what does precise-commits give us as a pre-commit over just running something like 'yarn prettier --check'

BenTopping (approved):
Happy to approve but maybe have a look into a few things:

- In some repos like traction-ui they are using lefthook as a pre-commit library? Not sure its necessary to change from husky but it might be bettter
- Maybe we should think about adding some prettier config so prettier only makes changes we want -  an example from traction here https://github.com/sanger/traction-ui/blob/develop/.prettierrc

package.json
seenanair: Sorry, I thought I committed changes to Readme already, but instead it was shelved.Now, I have pushed the changes to Readme file. Feel free to edit if you need any changes.

seenanair:
> Happy to approve but maybe have a look into a few things:
> 
> * In some repos like traction-ui they are using lefthook as a pre-commit library? Not sure its necessary to change from husky but it might be bettter
> * Maybe we should think about adding some prettier config so prettier only makes changes we want -  an example from traction here https://github.com/sanger/traction-ui/blob/develop/.prettierrc

Yes, I feel those are worth looking into. I will look into those and make necessary changes. 👍

seenanair:
@BenTopping I have added .nvmrc file and added the checks in .yml files for  ensuring node version specified in build . Please check.

BenTopping (approved):

PR #173, opened 2022-07-27
x889 - Search by work number

BenTopping:
- Added SGP Number option to search page


BenTopping:
Note: Codegen kept wanting to change a lot of things within the sdk file which is why the changes look high. Think it is just trying to arrange them in alphabetical order.

src/pages/Search.tsx
```diff
@@ -276,6 +298,13 @@ const columns: Column<SearchResultTableEntry>[] = [
     Header: "Labware Type",
     accessor: "labwareType",
   },
+  {
+    Header: "SGP Numbers",
+    Cell: (props: Cell<SearchResultTableEntry>) => {
+      const workNumbers = props.row.original.workNumbers;
+      return workNumbers.join(", ");
+    },
+  },
```
seenanair, line 307: I think it is better to have this column as sortable because there was a previous requirement to make all columns in search results to be sortable

seenanair (commented):
Looking good

src/pages/Search.tsx
BenTopping: @seenanair Its going to be a little tricky to sort because worknumbers column can have multiple work numbers, so not sure what the best way to sort is? I can apply the same sort as location uses maybe?

seenanair: @BenTopping Worknumber sort is implemented  in Dashboard page table. Would you please check that out?

PR #175, opened 2022-07-29
x904 - Error handling 

BenTopping:
Tightened up some tests, fix json block error messages with labware machine


src/lib/machines/labware/labwareMachine.ts
```diff
@@ -424,10 +425,7 @@ export const createLabwareMachine = (
           ) {
             return;
           }
-          const matchResult = e.data.message.match(/^.*\s:\s(.*)$/);
-          if (matchResult && matchResult.length > 1) {
-            ctx.errorMessage = matchResult[1];
-          }
+          ctx.errorMessage = handleFindError(e.data);
```
seenanair, line 428: @BenTopping With this, server error extraction is now based on extractServerErrors method, right?

seenanair (approved):

src/lib/machines/labware/labwareMachine.ts
BenTopping: Yep. HandleFindError calls extractServerErrors

PR #178, opened 2022-08-03
X915 enforce prettier and linting

BenTopping:
- Added lefthook to handle pre-commit checks
- Added eslint and prettier config
- Prettier'd and linted entire repo
- Updated readme
- Updated github actions to include prettier and lint check

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,13 @@
 node_modules
 public
+src/types/*
+src/serviceWorker.ts
+src/index.tsx
+cypress/support/*
+cypress/plugins/*
+cypress.config.ts
+CHANGELOG.md
+README.md
+.github
+.storybook
+.prettierignore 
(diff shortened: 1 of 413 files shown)


.prettierignore
```diff
@@ -1,2 +1,13 @@
 node_modules
 public
+src/types/*
+src/serviceWorker.ts
```
seenanair, line 4: Why this file is not included?

package.json
```diff
@@ -109,14 +110,14 @@
     "babel-plugin-polyfill-corejs2": "^0.3.1",
     "cypress": "^10.3.1",
     "cypress-intellij-reporter": "^0.0.7",
+    "eslint-config-prettier": "^8.5.0",
```
seenanair, line 113: This package is to manage the formatting rules to avoid conflicts with prettier, I suppose. Is it not required to install eslint package separately, if we install this?

seenanair (approved):
Looks good, great job. 
Just added a few doubts, please check.

.prettierignore
BenTopping: I was under the impression serviceWorker.ts was a generated file by another package so didnt want to touch it. Happy to remove it thought if you think it should be checked.

seenanair: Ohh, ok.In that case I think we can leave it.

PR #180, opened 2022-08-04
Removed warnings 

seenanair:
Removed XState machine rendering warnings , except 
- "Xstate useSelector - the result of getSnapshot should be cached to avoid an infinite loop" warning. This is an internal XState issue , discussed here [
](https://github.com/statelyai/xstate/discussions/3379)
- Switching between controlled and uncontrolled state for spinners which need to display an empty value in the beginning


src/components/DataFetcher.tsx
```diff
@@ -21,13 +21,14 @@ const DataFetcher = <E extends unknown>({
   dataFetcher,
   children,
 }: DataFetcherProps<E>) => {
-  const [state, send] = useMachine(
-    createDataFetcherMachine({
+  const dataFetcherMachine = React.useMemo(() => {
```
BenTopping, line 24: Is the general idea that useMemo stops the machine from constantly running?

src/components/forms/Input.tsx
```diff
@@ -5,7 +5,7 @@ import { FormikErrorMessage } from "./index";
 import classNames from "classnames";
 
 const defaultInputClassNames =
-  "focus:ring-sdb-100 focus:border-sdb-100 block border-gray-300 rounded-md disabled:opacity-75 disabled:cursor-not-allowed";
```
BenTopping, line 8: I think you added block when you added sanger-ui styling package. Maybe devel needs to be merged into this branch?

BenTopping (changes requested):
Will test locally once devel has been merged in and conflicts are fixed. Otherwise going to get some weird behaviour as this might have old packages etc.

src/components/DataFetcher.tsx
seenanair: The warning was "Machine given to `useMachine` has changed between renders. This is not supported and might lead to unexpected results.
Please make sure that you pass the same Machine as argument each time."
This prevents from giving a different instance of machine to useMachine on each render unless there is a change in required parameters. Please refer [https://xstate.js.org/docs/tutorials/reddit.html#splitting-machines](https://xstate.js.org/docs/tutorials/reddit.html#splitting-machines)

src/components/forms/Input.tsx
seenanair: Yes, I will resolve conflicts soon

BenTopping (approved):
Looks good, nice to have a clear console for future!

PR #181, opened 2022-08-08
x892 Highlight barcode on history page

### LINKED ISSUE #196: x892 - Highlight barcode in history page
**User story**
As a PO (Minal) I would like the barcode in the history page to be highlighted to make it quick and easy to find the barcode in the history

**Who are the primary contacts for this story**
Minal

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] if searching by barcode - highlight the relevant barcode in the history 
- [ ] Make text background yellow for search term barcode in the result columns

BenTopping:
Closes #196 

- Added background colour to matching barcodes on history page

### DIFF
--- a/src/components/history/History.tsx
+++ b/src/components/history/History.tsx
@@ -47,11 +47,10 @@ export default function History(props: HistoryProps) {
         accessor: 'sourceBarcode',
         Cell: (props: Cell<HistoryTableEntry>) => {
           const barcode = props.row.original.sourceBarcode;
-          let classes = '';
-          if (historyProps.kind === 'labwareBarcode' && barcode === historyProps.value) {
-            classes =
-              'bg-yellow-400 text-sp-600 hover:text-sp-700 font-semibold hover:underline text-base tracking-wide';
-          }
+          let classes =
+            historyProps.kind === 'labwareBarcode' && barcode === historyProps.value
+              ? 'bg-yellow-400 text-sp-600 hover:text-sp-700 font-semibold hover:underline text-base tracking-wide'
+              : '';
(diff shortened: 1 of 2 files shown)


src/components/history/History.tsx
```diff
@@ -24,79 +24,92 @@ export default function History(props: HistoryProps) {
...
+      },
+      {
+        Header: 'Source',
+        accessor: 'sourceBarcode',
+        Cell: (props: Cell<HistoryTableEntry>) => {
+          const barcode = props.row.original.sourceBarcode;
+          let style = undefined;
+          if (historyProps.kind === 'labwareBarcode' && barcode === historyProps.value) {
+            style = { backgroundColor: 'yellow' };
+          }
```
seenanair, line 53: Just a thought - may be a 'padding' as well to make the highlight bigger/obvious. I don't know whether it looks good, but may be worth a try

seenanair (commented):
Looks good 👍 
Do we need a test case for this highlight?

src/components/history/History.tsx
BenTopping: Tested with padding and it makes the positioning different a bit different to the non-highlighted barcodes so it looks a bit weird.

khelwood (commented):
Looks good, but the linter has some suggestions that look reasonable

khelwood (approved):

BenTopping:
> Looks good, but the linter has some suggestions that look reasonable

Yep, fixed those lint suggestions in another PR https://github.com/sanger/stan-client/pull/183

PR #182, opened 2022-08-09
Summary dashboard implementation

### LINKED ISSUE #200: x891 - Summary dashboard page
**User story**
As a PO (Minal), I want a summary dashboard page in Stan so the researchers and senior management can see works status and volume quickly and easily to reduce manually extracting data from the current detailed dashboard

**Who are the primary contacts for this story**
Mina

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] Table should show: Work type, status, num of work requests, total number of blocks / slides in work requests

**Additional context**
* Obvious link from home page
* Should have its own URL

seenanair:
Closes #200 
Separate page for summary dashboard accessible from //work_progress_summary
Link from Dashboard page to summary


src/components/StyledLink.tsx
```diff
@@ -1,12 +1,14 @@
...
 import { Link, LinkProps } from 'react-router-dom';
+import classNames from 'classnames';
 
 type StyledLinkProps = LinkProps & React.RefAttributes<HTMLAnchorElement>;
 
 const StyledLink: React.FC<StyledLinkProps> = ({ children, className, ...rest }) => {
-  const linkClassName =
-    className ?? 'text-sp-600 hover:text-sp-700 font-semibold hover:underline text-base tracking-wide';
-
+  const linkClassName = classNames(
```
BenTopping, line 8: what does the classNames method do over something like this? linkClassName = "text-sp-600 hover ... " + className. 
Also this change removes the flexibility of styling the link as you cant override the defaults. Not sure this is an issue if we want all StyledLinks to look the same

src/pages/WorkProgress.tsx
```diff
@@ -161,10 +162,24 @@ const WorkProgress = ({ workTypes }: { workTypes: string[] }) => {
...
+        </div>
       </AppShell.Header>
       <AppShell.Main>
         <div className="mx-auto">
+          <div className="mx-auto max-w-screen-lg mt-2 my-6 rounded-md space-y-4">
+            <div className={'flex flex-col w-full border border-gray-200 bg-gray-100 p-4 gap-y-4 text-lg'}>
+              <Heading level={3} showBorder={false}>
+                Summary Dashboard
+              </Heading>
+              <div className={' mx-auto flex w-full p-4 rounded-md justify-center bg-gray-200'}>
```
BenTopping, line 176: remove initial space in className

src/pages/WorkProgress.tsx
```diff
@@ -161,10 +162,24 @@ const WorkProgress = ({ workTypes }: { workTypes: string[] }) => {
   return (
     <AppShell>
       <AppShell.Header>
-        <AppShell.Title>STAN</AppShell.Title>
+        <div className="flex flex-row">
```
BenTopping, line 165: Assume this is something needed so the summary dashboard doesnt get too close to the header, just checking if its actually needed or not

seenanair: In fact, this is not required. I have removed it. Good finding 👍

src/components/StyledLink.tsx
seenanair: The code seems bit different in mine _**const linkClassName = classNames(
    'text-sp-600 hover:text-sp-700 font-semibold hover:underline text-base tracking-wide',
    className
  );**_

seenanair: The end result is same as you specified but I think  it does few more extra checks. The classNames function takes any number of arguments which can be a string or object. For example, If the value associated with a given key is falsy, that key won't be included in the output. So, it is more like a convenience method especially  if there are multiple strings.

BenTopping: Ah ok. Sorry I poorly explained it, my code looks the same as yours but I was just wondering the advantages of classNames over just extending the string, but as you've just explained its fine.

BenTopping (approved):

src/types/stan.ts
```diff
@@ -297,12 +297,13 @@ export type HistoryProps = HistoryStrProps | HistoryNumProps;
  */
 export const statusSort = (rowAStatus: WorkStatus, rowBStatus: WorkStatus) => {
   const statusArray: WorkStatus[] = [
+    WorkStatus.Unstarted,
```
BenTopping, line 300: These changes should be reverted as this is the method used for standard dashboard and needs a particular order.

src/pages/WorkProgressSummary.tsx
```diff
@@ -0,0 +1,153 @@
...
+    accessor: 'workType',
+    sortType: (rowA, rowB) => {
+      return alphaNumericSortDefault(rowA.original.workType, rowB.original.workType);
+    }
+  },
+  {
+    Header: 'Status',
+    accessor: 'status',
+    sortType: (rowA, rowB) => {
+      return statusSort(rowA.original.status, rowB.original.status);
```
BenTopping, line 134: Change here to use summaryStatusSort or similar

BenTopping (changes requested):
Comments above need changing to avoid clashing with normal dashboard status sort

BenTopping (approved):
Nice 👍

PR #212, opened 2022-09-14
Display 'Add Identical Tissue' button only in Original sample registration page

seenanair:


src/pages/registration/RegistrationForm.tsx
```diff
@@ -356,24 +356,29 @@ const RegistrationForm = <T extends TissueValues<B>, B>({
...
-                    setFieldValue(`tissues[${currentIndex + 1}].blocks[${indx}].replicateNumber`, '');
-                    setFieldValue(`tissues[${currentIndex + 1}].blocks[${indx}].externalIdentifier`, '');
-                    setFieldValue(
-                      `tissues[${currentIndex + 1}].blocks[${indx}].spatialLocation`,
-                      values.tissues[currentIndex].blocks[indx].spatialLocation
-                    );
-                  });
+                //Only add identical tissue button for original sample registration
+                onIdenticalTissueButton={
+                  'solution' in values.tissues[0].blocks[0]
```
khelwood, line 361: What does `'solution' in values.tissues[0].blocks[0]` represent?

seenanair: solution field only exists for original sample registration, so it is used here to distinguish between block registration and original sample registration

PR #217, opened 2022-09-28
X941 - Visium perm without stain

BenTopping:
Visium permeabilisation can now occur without stained labware. 
These changes check that if the labware is stained and show a warning popup if they are unstained.

### DIFF
--- a/src/pages/VisiumPerm.tsx
+++ b/src/pages/VisiumPerm.tsx
@@ -60,21 +60,24 @@ export default function VisiumPerm() {
   const [warnBeforeSave, setWarnBeforeSave] = React.useState(false);
   const { serverError } = current.context;
 
-  const onSubmit = async (values : RecordPermRequest) => {
-    const isLabwareStained = await stanCore.GetLabwareOperations({
-      barcode: values.barcode, operationType: "Stain"
-    }).then(r => r.labwareOperations?.length);
-  
+  const onSubmit = async (values: RecordPermRequest) => {
+    const isLabwareStained = await stanCore
+      .GetLabwareOperations({
+        barcode: values.barcode,
+        operationType: 'Stain'
+      })
+      .then((r) => r.labwareOperations?.length);
+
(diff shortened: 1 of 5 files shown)


src/pages/VisiumPerm.tsx
```diff
@@ -54,11 +55,29 @@ export default function VisiumPerm() {
...
     });
   }, []);
-  const [current, send] = useMachine(formMachine);
 
+  const [current, send] = useMachine(formMachine);
+  const [warnBeforeSave, setWarnBeforeSave] = React.useState(false);
   const { serverError } = current.context;
 
   const onSubmit = async (values: RecordPermRequest) => {
+    const isLabwareStained = await stanCore
```
BenTopping, line 64: only concern is if this request fails (which in theory it should never fail), users wont be alerted and the check wont work. Not sure this is a big deal though.

src/pages/VisiumPerm.tsx
```diff
@@ -54,11 +55,29 @@ export default function VisiumPerm() {
...
+        barcode: values.barcode,
+        operationType: 'Stain'
+      })
+      .then((r) => r.labwareOperations?.length);
+
+    if (!!isLabwareStained) {
+      handleSave(values);
+    } else {
+      setWarnBeforeSave(true);
+    }
```
seenanair, line 75: Just a question - Would it be more appropriate to do it as an xState service while scanning the labware itself than doing it on 'Save operation'?

BenTopping: hmm good question. I guess that could be done by showing the pop up modal after the labware has been scanned if it doesnt have stains. I think i did it this way because in other pages like slotCopy(visium cDna) the warning modal shows up on submit

seenanair: It's ok. We can go with this as it is doing it like other pages 👍

seenanair (approved):

PR #229, opened 2022-10-24
X937 original sample page update

seenanair:
'Fixative' and 'Solution' columns added in confirmation page


src/pages/registration/RegistrationForm.tsx
```diff
@@ -158,7 +158,7 @@ const RegistrationForm = <T extends TissueValues<B>, B>({
                     return (
                       <motion.div
                         ref={blockIndex === values.tissues[currentIndex].blocks.length - 1 ? lastBlockRef : null}
-                        key={block.clientId}
+                        key={blockIndex}
```
BenTopping, line 161: better option for key than an index?

seenanair: Unfortunately, in this context there is not a key identifier possible, all the fields contained in blocks can repeat including clientId

src/pages/OriginalSampleRegistration.tsx
```diff
@@ -146,15 +115,124 @@ function OriginalSampleRegistration({ registrationInfo }: RegistrationParams) {
...
+    ];
+  }, []);
+
+  const formatSuccessData = React.useCallback(
+    (labware: LabwareFieldsFragment[]) => {
+      return labware.map((lw, indx) => {
+        return {
+          labware: lw,
+          extraData:
+            sampleRegistrationData && sampleRegistrationData.length > indx
```
BenTopping, line 228: not 100% on what this is doing. Is extra data only ever the solution from sampleRegistrationData? If so does it need to be called extraData, could just be called solution. Also it could be flakey grabbing a value based on a position in an array, is there a better way to link labware to the sampleRegistration data?

src/pages/OriginalSampleRegistration.tsx
```diff
@@ -146,15 +115,124 @@ function OriginalSampleRegistration({ registrationInfo }: RegistrationParams) {
...
+        Header: 'Spatial location',
+        id: 'spatialLocation',
+        accessor: (result: LabwareResultData) =>
+          valueFromSamples(result.labware, (sample: SampleFieldsFragment) => String(sample.tissue.spatialLocation.code))
+      },
+      {
+        Header: 'Replicate',
+        id: 'replicate',
+        accessor: (result: LabwareResultData) =>
+          valueFromSamples(result.labware, (sample: SampleFieldsFragment) => String(sample.tissue.replicate))
```
BenTopping, line 205: can sample.tissue.replicate be undefined here? does it need a `?? ""`

src/pages/registration/SummaryBox.tsx
```diff
@@ -97,7 +97,7 @@ const SummaryBox = ({
                 e.preventDefault();
                 setCurrentFormIndex(tissueIndex);
               }}
-              key={tissue.clientId}
+              key={tissueIndex}
```
BenTopping, line 100: better option for key than an index?

src/pages/registration/RegistrationForm.tsx
BenTopping: Ah ok, thats fine then

src/pages/registration/RegistrationSuccess.tsx
```diff
@@ -13,12 +13,37 @@ import AppShell from '../../components/AppShell';
...
+  columns: Column<T>[];
+};
+
+const RegistrationSuccess = <T extends Required<LabwareContainType> | LabwareFieldsFragment>({
+  successData,
+  columns
+}: RegistrationSuccessProps<T>) => {
+  /**Memoised labware from successData**/
+  const labware = React.useMemo(() => {
+    /**Is Result of type LabwareFieldsFragment**/
```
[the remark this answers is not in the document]
seenanair: typeof only checks against number,boolea, string,null symbol,function etc . But, here we need to check against whether the given prop is an instance of a particular object type which I believe is not possible through typeof

src/pages/OriginalSampleRegistration.tsx
seenanair: Yes, replicate can be undefined in case of original sample registration

src/pages/registration/SummaryBox.tsx
seenanair: Again , same as before there is not a unique id field exists for a given data structure

src/pages/OriginalSampleRegistration.tsx
seenanair: As it is template structure, I thought that will give flexibility to provide any other extra data that we need to disaply other than labware, if such cases arise in future realted to other registrations as well.I know, the ideal would have been to return solution as part of mutation result. But another

src/pages/registration/RegistrationSuccess.tsx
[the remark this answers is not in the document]
BenTopping: Yes you are right. Just checked https://stackoverflow.com/questions/44078205/how-to-check-the-object-type-on-runtime-in-typescript

src/pages/OriginalSampleRegistration.tsx
BenTopping: Ok thats fair enough. Im not too sure on this use of index though. It relies on the labware and sampleRegistration being the same length and order

src/pages/registration/RegistrationSuccess.tsx
```diff
@@ -36,15 +61,15 @@ const RegistrationSuccess: React.FC<RegistrationSuccessProps> = ({ labware, colu
...
+            <DataTable columns={columns} data={successData} />
           </motion.div>
 
           <div className="flex flex-row items-center sm:justify-end">
             <motion.div
               variants={variants.fadeInWithLift}
               className="sm:max-w-xl w-full border-gray-200 p-4 rounded-md bg-gray-100 shadow"
             >
-              <LabelPrinter labwares={labware} />
+              {<LabelPrinter labwares={labware} />}
```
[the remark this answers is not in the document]
seenanair: No, I will remove it 👍

src/pages/OriginalSampleRegistration.tsx
seenanair: Yes, I have asked David whether it is possible to return the 'solution' as part of the mutation result from which we are getting labware information.

BenTopping: Ah ok sorry I misunderstood 👍

seenanair: It's ok. Again, do you an any other appropriate name for 'extraData' field?

BenTopping: I think extraData is probably fine because its a broad range of data so its not specific. Maybe registrationData or something but up to you.

src/components/originalSampleProcessing/blockProcessing/BlockProcessing.tsx
```diff
@@ -20,7 +20,7 @@ import Planner, { PlanChangedProps } from '../../planning/Planner';
 import createFormMachine from '../../../lib/machines/form/formMachine';
 import { stanCore } from '../../../lib/sdk';
 import WorkNumberSelect from '../../WorkNumberSelect';
-import columns from '../../dataTable/labwareColumns';
+import columns from '../../dataTableColumns/labwareColumns';
```
BenTopping, line 23: multiple imports from the same file, thought prettier would have picked it up. Maybe change to `import { Columns, LabwareScanTableColumns } from '../../dataTableColumns/labwareColumns'`

BenTopping: Think this happens in a few places

src/pages/OriginalSampleRegistration.tsx
```diff
@@ -146,15 +116,117 @@ function OriginalSampleRegistration({ registrationInfo }: RegistrationParams) {
...
+      }
+    ];
+  }, []);
+
+  const formatSuccessData = React.useCallback((registerResult: RegisterResultFieldsFragment) => {
+    return registerResult.labware.map((lw) => {
+      return {
+        labware: lw,
+        extraData: [
+          registerResult.labwareSolutions.find((lwSolution) => lwSolution?.barcode === lw.barcode)?.solutionName ?? ''
```
BenTopping, line 225: Nice 👍

BenTopping (approved):
Few minor changes but otherwise looks good

src/components/originalSampleProcessing/blockProcessing/BlockProcessing.tsx
seenanair: Good finding!

src/components/originalSampleProcessing/potProcessing/PotProcessing.tsx
```diff
@@ -25,7 +25,7 @@ import { optionValues } from '../../forms';
 import FormikInput from '../../forms/Input';
 import PotProcessingLabwarePlan from './PotProcessingLabwarePlan';
 import ProcessingSuccess from '../ProcessingSuccess';
-import columns from '../../dataTable/labwareColumns';
+import columns from '../../dataTableColumns/labwareColumns';
```
BenTopping, line 28: Duplicate imports here as well

seenanair: Done 👍

BenTopping (approved):

PR #230, opened 2022-10-24
Changed Dual Index Plate name

seenanair:
Dual Index Plate names changed to Fresh Frozen - Dual Index TT Set A and
FFPE - Dual Index TS Set A


BenTopping (approved):
Looks good. Just wondering if we need a test using 'FFPE - Dual Index TS Set A' plate type. Does the plate type affect any logic?

seenanair:
> Looks good. Just wondering if we need a test using 'FFPE - Dual Index TS Set A' plate type. Does the plate type affect any logic?

Not actually on client side. It is just another field to pass on to the mutation.

seenanair:
@BenTopping  But another thought, while you are here, I will give it a go with unit test cases for this story

tests/unit/components/slotMapper/reagentTransferSlotMapper.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+const destLabware = createLabware('STAN-5111');
+
+describe('ReagentTransferSlotMapper.tsx', () => {
+  it('displays the parent div elements for source and destination', () => {
+    const reagentSlotMapper = render(
+      <ReagentTransferSlotMapper initialSourceLabware={undefined} initialDestLabware={undefined} />
+    );
+
+    //Displays div element for source labware
+    const sourceLabwareElem = getById(reagentSlotMapper.container, 'sourceLabwares');
```
BenTopping, line 21: You could add a `data-testid=sourceLabwares` to the reagentTransferSlotMapper components to use as a unique identifier for the tests. Then get the component by its testId in the tests like c`onst sourceLabwareElem = screen.getByTestId('sourceLabwares')`

tests/unit/components/slotMapper/reagentTransferSlotMapper.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+enableMapSet();
+afterEach(() => {
+  cleanup();
+});
+const srcLabware = createLabware('123456789123456789012345');
+const destLabware = createLabware('STAN-5111');
+
+describe('ReagentTransferSlotMapper.tsx', () => {
+  it('displays the parent div elements for source and destination', () => {
+    const reagentSlotMapper = render(
```
BenTopping, line 16: not sure assigning it a variable is necessary as you can reference the rendered element through`screen`

PR #231, opened 2022-10-25
x938 - Changes to fix cypress test flaking

### LINKED ISSUE #227: x938 Cypess CI testing when run as GitHub
User Story
As the Dev Team we would like to investigate the Cypress CI testing when run as a GitHub action to identify reasons for the test to fail so regularly

Acceptance
* Investigate the GitHub Action setup to see if it can be improved.
* Talk to other members of the team OR search for other sources to see if this is a common issue.
* Investigate if internal test logic makes timing assumptions?

BenTopping:
Closes #227 

Previously cypress was running off electron, in theory this could have caused the difference in github CI to local tests as locally we use chrome.

Running off chrome appears to make the tests 1-2 mins slower.

In future we may need to upgrade the docker image to the latest versions of chrome using https://github.com/cypress-io/cypress-docker-images

Flakiness is difficult to test but so far this PR has **passed 5/5** re-runs of cypress tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
   cypress:
     name: Run cypress tests
     runs-on: ubuntu-latest
-    container: cypress/included:9.4.1
+    container: cypress/browsers:node16.17.1-chrome106-ff105-edge
     steps:
       - uses: actions/checkout@v2
         name: Checkout
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,9 @@ jobs:
   cypress:
     name: Run cypress tests
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -13,6 +13,9 @@ jobs:
   cypress:
     name: Run cypress tests
     runs-on: ubuntu-latest
+    container: 
+      image: cypress/browsers:node16.17.1-chrome106-ff105-edge
+      options: --user 1001
```
seenanair, line 18: What is --user 1001 indicates?

BenTopping: There was a permissions issue with the docker container. [This discussion here](https://github.com/cypress-io/github-action/issues/446#issuecomment-987015822) outlines the issue and temporary fix.

seenanair (approved):

PR #234, opened 2022-10-27
Unit test cases

seenanair:
Unit test cases added for WorkNumberSelect

### DIFF
--- a/src/components/WorkNumberSelect.tsx
+++ b/src/components/WorkNumberSelect.tsx
@@ -66,6 +66,9 @@ export default function WorkNumberSelect({ name, label, workNumber, onWorkNumber
       );
     }
     fetchActiveWorkNumbers();
+    return () => {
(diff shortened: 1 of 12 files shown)


src/components/WorkNumberSelect.tsx
```diff
@@ -88,13 +91,14 @@ export default function WorkNumberSelect({ name, label, workNumber, onWorkNumber
   );
 
   const validateWorkNumber = () => {
+    console.log('worknumber=' + workNumber);
```
BenTopping, line 94: Can get rid of these console logs now

BenTopping (approved):
Remove console logs and then should be fine. Good job 👍

src/components/WorkNumberSelect.tsx
seenanair: 👍

PR #249, opened 2022-11-29
File viewing error fix

seenanair:
The router path to redirect to viewing page for non-authenticated user is added.

### DIFF
--- a/src/components/Routes.tsx
+++ b/src/components/Routes.tsx
@@ -383,6 +383,27 @@ export function Routes() {
           );
         }}
       />
+      <Route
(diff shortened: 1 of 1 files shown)


khelwood:
It looks like this was clobbered in commit `4460e282fae113`(Tube section registration), presumably in a merge conflict.

seenanair:
> It looks like this was clobbered in commit `4460e282fae113`(Tube section registration), presumably in a merge conflict. Could you check nothing else was affected in this way?

I checked using git diff and this is the only one listed.

PR #264, opened 2022-12-15
Visium QC update - reagent lot number

seenanair:


seenanair:
@khelwood  Fix for the problem Katy reported for x957 - " it should be Reagent LOT number not Slide LOT number. And I keep getting an error of "failed to record visium QC on the page" not sure what I have done wrong?!".

khelwood (approved):

PR #278, opened 2023-01-17
X967 unit testcases

seenanair:
Unit test cases added for 
- Slot, SlotColumnInfo, labware.machine


khelwood (commented):
Do these new tests replace some cypress tests that can be deleted?

seenanair:
> Do these new tests replace some cypress tests that can be deleted?

I need to look into that as the next step.

khelwood (approved):

PR #289, opened 2023-02-09
X979 transfer stored labware

seenanair:
Transferring items in a location to another
  - Copy button next to store location details
  - Transfer items section for location with no children
  - Text box to enter/paste source location barcode
  - On enter key, the description will be displayed if valid, otherwise displays error
  - Clicking on 'Transfer' button transfers stored items from source to currently displayed location
  - Displays success/ error message indicating output of transfer action
  - On successful transfer, the current location is updated with moved items

### DIFF
--- a/src/lib/services/locationService.ts
+++ b/src/lib/services/locationService.ts
@@ -123,3 +123,29 @@ export async function setLocationCustomName(
 
   return setLocationCustomName;
 }
+
+/**
+ * Get details of specified location
+ */
+export async function getLocation(barcode: string): Promise<LocationFieldsFragment | undefined> {
+  const { location } = await stanCore.FindLocationByBarcode({ barcode });
+  if (!location) {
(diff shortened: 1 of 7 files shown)


src/lib/services/locationService.ts
```diff
@@ -123,3 +123,29 @@ export async function setLocationCustomName(
...
+ */
+export async function transferItems(
+  sourceBarcode: string,
+  destinationBarcode: string
+): Promise<LocationFieldsFragment | undefined> {
+  try {
+    const { transfer } = await stanCore.TransferLocationItems({ sourceBarcode, destinationBarcode });
+    return transfer;
+  } catch (e) {
+    throw e;
```
khelwood, line 149: What is the reason to catch e and then throw e? Why is that better than not catching it?

khelwood (approved):

PR #294, opened 2023-02-28
X980 worknumber for released labware

seenanair:
- Can select a worknumber for each released labware
- 'Select all' work number option
- Worknumber is optional
- Displays the worknumber assigned to the labware as the default value on scan


src/graphql/queries/GetSuggestedWorkForLabware.graphql
```diff
@@ -0,0 +1,10 @@
+query GetSuggestedWorkForLabware($barcodes: [String!]!) {
+  suggestedWorkForLabware(barcodes: $barcodes) {
+    suggestedWorks {
+      ...SuggestedWorkFields
+    }
+    works {
```
khelwood, line 6: I don't think you need to get the `works` stuff here at all: as far as I can tell you only use the `suggestedWorks`.

seenanair: Updated. Thank you

src/mocks/handlers/labwareHandlers.ts
```diff
@@ -93,6 +96,24 @@ const labwareHandlers = [
...
+    (req, res, ctx) => {
+      const work = workFactory.build({ workNumber: 'SGP1008' });
+      return res(
+        ctx.data({
+          suggestedWorkForLabware: {
+            suggestedWorks: [...req.variables.barcodes].map((barcode) => ({
+              workNumber: work.workNumber,
+              barcode: barcode
+            })),
+            works: [work]
```
khelwood, line 112: I think `works:` isn't needed, in line with the last change

seenanair: Yes, updated

khelwood (approved):

PR #307, opened 2023-03-21
Spatial location bug fix

seenanair:
Spatial location field error while registering multiple tissues

How to test :-  
 - Enter details for tissue
- Use either 'Add Another Tissue' or 'Add Another Tissue Block' 
- Should display an empty value in 'Spatial location' field. 
- Enter all required fields
- Press 'Register' button


seenanair:
@khelwood 
- The branch is created from rc.

 Please test the registration as well when you get time just to make sure, there aren't other issues.
It seems the following scenario created the bug

- Enter details for tissue
- Use either 'Add Another Tissue' or 'Add Another Tissue Block'
- Should display an empty value in 'Spatial location' field.
- Enter all required fields
- Press 'Register' button

khelwood (approved):

PR #327, opened 2023-04-27
Cypress test case failure test on CI run

seenanair:


cypress/e2e/pages/location.cy.ts
```diff
@@ -142,15 +142,20 @@ describe('Location', () => {
     });
 
     context('when back button is pressed in browser', () => {
```
khelwood, line 144: This context says "back button is pressed". Does `cy.visit('/locations/STO-002');` match that description?

cypress/e2e/pages/location.cy.ts
```diff
@@ -142,15 +142,20 @@ describe('Location', () => {
...
-        cy.go('back');
         cy.findByText('STAN-2111').should('exist');
         cy.findByText('STAN-4111').should('exist');
         cy.findByText('STAN-3111').should('not.exist');
       });
     });
 
     context('when refreshing the page', () => {
+      before(() => {
+        cy.reload();
```
khelwood, line 157: Does this need to reload in both the before() and subsequently in the `it('should...)` block?

cypress/e2e/pages/location.cy.ts
seenanair: Yes, the following code shows how it was navigated. The intention of this test case is to  ensure 'awaitingLabware' list on display is updated following an add action on an item  (which is already in awaiting storage) and it is reflected on navigating to another page. So, even if it is not an exact back location, it make sense to any other location, I hope.
      _cy.visit('/locations/STO-002');
      cy.findByText('Rack 1 in Freezer 1 in Room 1234').click();
      cy.findByText('Box 1 in Rack 1 in Freezer 1 in Room 1234').click();_

khelwood: OK as long as you're happy it's testing what it says :+1:

seenanair: I have changed the test case description to _'when navigating to a previous location or another page'_ . Hope that ids makes it more clear

khelwood (approved):

PR #331, opened 2023-05-04
Snyk github action testing

seenanair:
Closes #

*
*
* ...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,7 +51,8 @@ jobs:
         continue-on-error: true # To make sure that SARIF upload gets called
         with:
           command: code test
-          args: --sarif-file-output=snyk.sarif
+          args: --sarif-file-output=snyk.sarif --severity-threshold=high
+
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
       - name: Upload result to GitHub Code Scanning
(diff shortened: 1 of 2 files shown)


khelwood (commented):
Looks good but could you give me a brief explanation?

seenanair:
> Looks good but could you give me a brief explanation?

Yes, sure.  But it still uses SNYK_TOKEN from my account which need to be changed to a service account. I couldn't set up a service account as I didn't have admin access rights in Snyk.

The configuration was as below -
1. Copied the API_TOKEN from my account (which ideally need to be service account) in Snyk
2. Create  and add SNYK_TOKEN variable in stan-client project (Settings -> Secrets and variables -> Repository secretes) with about value

Then added github actions required
  - snyk/actions/node@master is a pre-created container available (https://docs.snyk.io/integrations/ci-cd-integrations/github-actions-integration/snyk-node-action)
     I believe it runs , 'snyk test' by default
      'command' - to give specific  snyk commands if need to be changed in our case we use 'code test' to avoid dependency scanning
      args -  
                  --severity-threshold=high    - will only look for high severity warnings
                  --sarif-file-output=snyk.sarif  - to create an output file for results

 continue-on-error: true - To avoid blocking CI run in case  snyk test fails

The following part is to upload Snyk scan results to GitHub Code Scanning 
     _- name: Upload result to GitHub Code Scanning
        uses: github/codeql-action/upload-sarif@v2
        with:
          sarif_file: snyk.sarif_

But this is not yet working. I assume it only works for 'monitor' command which I am trying to check now.

Hope this makes sense.

khelwood (approved):

PR #341, opened 2023-05-25
X1034 dep update

seenanair:
Dependency updates for some packages . The main ones are faker, msw, tailwind, typescript

- Changes associated with faker library  deprecation /change of many methods 
- Changes for TS - The generic type is not supported without constraint, so an extends keyword which clearly state what the user want a generic type is required


khelwood:
Is this actually ready to review? It is marked as a draft

seenanair:
> Is this actually ready to review? It is marked as a draft

Yes it is. Sorry, I will change the status

src/pages/Search.tsx
```diff
@@ -170,7 +170,7 @@ function Search({ searchInfo, urlParamsString }: SearchProps) {
 
   const numDisplayed = viewAllRecords ? searchResult?.numDisplayed : memoUniqueBarcodeData.length;
   const numRecords = viewAllRecords ? searchResult?.numRecords : memoUniqueBarcodeData.length;
-
+  debugger;
```
khelwood, line 173: debugger

seenanair:
> Happy to approve if the tests will actually pass

I will look into this issue

PR #359, opened 2023-07-03
X967 unit test added for slotmapper component

seenanair:


src/components/labware/Labware.tsx
```diff
@@ -144,13 +144,15 @@ const Labware = ({
   slotBuilder
 }: React.PropsWithChildren<LabwareProps>) => {
   const labwareMachine = React.useMemo(() => {
-    return createLabwareMachine({
+    return createLabwareMachine();
+  }, []);
+  const [current, send] = useMachine(labwareMachine, {
```
khelwood, line 149: Why has this context parameter been introduced?

src/components/labware/labware.machine.ts
```diff
@@ -1,34 +1,21 @@
...
-  SelectionMode
-} from './labware.types';
+import { LabwareMachineContext, LabwareMachineEvent, LabwareMachineSchema } from './labware.types';
 import { assign } from '@xstate/immer';
 import { pure } from 'xstate/lib/actions';
-import { SlotFieldsFragment } from '../../types/sdk';
 import { emptySlots, filledSlots, findSlotByAddress, isSlotEmpty, isSlotFilled } from '../../lib/helpers/slotHelper';
 import { sortDownRight } from '../../lib/helpers/labwareHelper';
 
-interface CreateLabwareMachineParams {
```
khelwood, line 15: Why has this interface been deleted?

src/components/labware/Labware.tsx
seenanair: There was a warning before in console 'Machine given to `useMachine` has changed between renders. This is not supported and might lead to unexpected results.'
So, if we pass in parameters while creating  machine based on a component prop, machine instance will be recreated whenever this prop/parameter changes and there by changing the useMachine which accept this machine instance as param which is not recommended. This  way of creating and using the  state machine , in which we create a static instance and then defining the context of state machine using the passed in parameters which will ensure that the machine instance will not be changed.

src/components/labware/labware.machine.ts
seenanair: This interface was to define the structure of params for createLabwareMachine which we are not using any more

khelwood (approved):

Issue #237, opened 2022-11-10
x934 As TS (Katy), I want to select a Program when creating a new work number on the SGP Management page, because this will allow us to track/filter by program type now we are running requests across Sanger.


KatyTudor:
"To be a dropdown list edited on the config page: CellGen, CASM, HUMGen, SGP
Add to the work_view
Projects are not part of programs.
Program is mandatory for work numbers (needs backfilling for old ones). Info for patch is [here](https://docs.google.com/spreadsheets/d/1op2DSEEneTAvaDfgnTDWAx8T11jTBsALZt-uTRtzdEM/edit?usp=sharing)

khelwood:
@KatyTudor 
Need programs for 
R&D17
R&D18
SGP63
SGP64

KatyTudor:
@khelwood I have added these to the google sheet now, they were created after I made the list sorry, might need to revise it just as we release the feature

KatyTudor:
@khelwood Can you also add PaM to the list sorry

PR #364, opened 2023-07-27
x1054

sabrine33:


src/components/Routes.tsx
```diff
@@ -354,6 +354,7 @@ export function Routes() {
               {(dataFetcher) => {
                 return (
                   <LabwareDetails
+                    samplePositionResults={dataFetcher.visiumPermData.samplePositionResults}
```
seenanair, line 357: I think we need to make use of _samplePositions_ query on core side to fetch samplePosition information related to a labware. _samplePositionResults_ are not related to visiumPermData

src/lib/services/historyService.ts
```diff
@@ -44,9 +49,19 @@ export async function findHistory(historyProps: HistoryProps): Promise<Array<His
 
   const labwareMap: Map<number, LabwareFieldsFragment> = new Map();
   const sampleMap: Map<number, SampleFieldsFragment> = new Map();
+  const samplePositionMapByOpId: Map<number, Map<number, SamplePositionFieldsFragment>> = new Map();
 
   history.labware.forEach((lw) => labwareMap.set(lw.id, lw));
   history.samples.forEach((sample) => sampleMap.set(sample.id, sample));
+  history.samplePositionResults.forEach((samplePosition) => {
```
seenanair, line 56: samplePositionResults are not associated with history data .Better to use  _samplePositions_ query on core side

src/mocks/handlers/historyHandlers.ts
```diff
@@ -41,7 +41,8 @@ export function buildHistory(workNumber?: string): HistoryFieldsFragment {
     __typename: 'History',
     samples: [sample],
     labware: [sourceLabware, destinationLabware].map(buildLabwareFragment),
-    entries
+    entries,
+    samplePositionResults: []
```
seenanair, line 45: This mocking is not correct as the core side doesn't support this

src/mocks/handlers/visiumHandlers.ts
```diff
@@ -53,9 +54,11 @@ const handlers = [
     }
 
     const labware = createLabware(barcode);
+    const samplePositionResults: SamplePositionFieldsFragment[] = [];
     return res(
       ctx.data({
         visiumPermData: {
+          samplePositionResults,
```
seenanair, line 61: This mocking is not correct as the core side doesn't support this

seenanair (commented):
I think some changes are required in the way the samplePosition data is acquired. Please ask if you have any further questions

src/components/Routes.tsx
sabrine33: I updated the core code accordingly, so the samplePositions is part of the visimuPermData object, unless you mean this is not logically correct, and we need to perform two separate query

src/lib/services/historyService.ts
sabrine33: I updated the core code accordingly, so the samplePositions is part of the History object, unless you mean this is not logically correct, and we need to perform two separate query

src/mocks/handlers/historyHandlers.ts
sabrine33: The core code, is been updated accordingly, you can have a look here [#237](https://github.com/sanger/stan-core/pull/237)

src/mocks/handlers/visiumHandlers.ts
sabrine33: The core code, is been updated accordingly, you can have a look here https://github.com/sanger/stan-core/pull/237

src/components/Routes.tsx
seenanair: Ohh, ok, In that case it is fine. But I couldn't see this update in src/graphql/queries/FindPermData.graphql.  I think that needs to updated as well

sabrine33: I did not update the FindPerData.graphql file as it is referring to the LabwareFields type, which already been updated.

seenanair (commented):
I think some changes are required in the way the samplePosition data is acquired. Please ask if you have any further questions

seenanair (approved):

khelwood:
@seenanair Can you please fix the conflicts so we can merge this in?

KatyTudor:
@sabrine33 sorry I think there's been a slight cross over on terminology, for the Slot ID instead of the information the back end STAN sees its the A1, B1, etc. value the team would need to see and would make sense to them. think its the slot address. Sorry its probably me not explaining it properly/getting terminology wrong

KatyTudor:
@sabrine33 the column titles are now correct but the information isn't be filled in on the history page (Picture 1 below) or the history on the labware details page (picture 2 below)

Screenshot 2023-09-01 135115

Screenshot 2023-09-01 135053

sabrine33:
Thank you katy, I think this is due to the section position being empty in this case, I will rectify it to make it work either way.

KatyTudor:
thanks @sabrine33

PR #370, opened 2023-08-03
X1013 block file reg

sabrine33:


src/pages/BlockRegistration.tsx
```diff
@@ -161,7 +184,38 @@ function BlockRegistration({ registrationInfo }: RegistrationParams) {
...
+      if (!isSuccess) return;
+      if (result && 'barcodes' in result) {
+        const barcodes: string[] = result['barcodes'];
+        const labwarePromises = barcodes.map((barcode: string) => stanCore.FindLabware({ barcode }));
+        Promise.all(labwarePromises)
+          .then((labwares) => {
+            if (labwares.length > 0) {
+              setFileRegisterResult(labwares.map((labware) => labware.labware!) as LabwareFieldsFragment[]);
+            } else {
+              displayWarningMsg(`No record has been registered. Please check your file.`);
```
seenanair, line 204: Would it be better to use 'block instead of 'record'?

src/pages/BlockRegistration.tsx
```diff
@@ -161,7 +184,38 @@ function BlockRegistration({ registrationInfo }: RegistrationParams) {
...
+            }
+          })
+          .catch(() => {
+            displayWarningMsg('Cannot retrieve details of newly registered block.');
```
seenanair, line 208: block(s) would be good

seenanair (approved):
Looking good 👍

PR #372, opened 2023-08-07
fixes X1060

sabrine33:


seenanair (commented):
Looks good.

seenanair:
> @seenanair, yes the layout is still the same, I am only hiding the measurement table until the user selects a measurement name, when the QC type is set visium concentration

Without modifying classes, if we hide a component I believe there might be a jump for labware from left to right once we show it. It would be good to avoid it by specifying justify for the labware as well as to centre it

seenanair (approved):

PR #377, opened 2023-08-17
X1062 probe hybridisation qc

sabrine33:


src/pages/ProbeHybridisationQC.tsx
```diff
@@ -0,0 +1,391 @@
...
+  });
+  return completionRequest;
+};
+
+const formInitialValues: ProbeHybridisationQCFormValues = {
+  workNumber: '',
+  labwares: {
+    '': {
+      globalComments: [],
+      completionDateTime: '',
```
seenanair, line 128: completionDateTime can be initialised to the current time

src/pages/ProbeHybridisationQC.tsx
```diff
@@ -0,0 +1,391 @@
...
+    return options.map((v) => v.value);
+  }, []);
+
+  const updateSectionCommentsFromGlobal = useCallback(
+    (options: OptionType[], labware: LabwareFieldsFragment, values, setFieldValue) => {
+      labware.slots.forEach((slot) => {
+        slot.samples.forEach((sample) => {
+          const sampleAddressId = `${slot.address}-${sample.id}`;
+          const oldSelected = values.labwares[labware.barcode]?.sampleAddressComments?.[sampleAddressId] || [];
+          const updatedSelected = [...oldSelected, ...mapCommentOptionsToValues(options)];
```
seenanair, line 202: Why we need oldSelected and updateSelected?

src/pages/ProbeHybridisationQC.tsx
```diff
@@ -0,0 +1,391 @@
...
+    );
+    await submitForm(values);
+  };
+
+  const mapCommentOptionsToValues = useCallback((options: OptionType[]) => {
+    return options.map((v) => v.value);
+  }, []);
+
+  const updateSectionCommentsFromGlobal = useCallback(
+    (options: OptionType[], labware: LabwareFieldsFragment, values, setFieldValue) => {
```
seenanair, line 197: It would be good to specify values ad setFieldValue as it defaults to 'any' in typescript which should generally be avoided whenever possible

src/pages/ProbeHybridisationQC.tsx
```diff
@@ -0,0 +1,391 @@
...
+
+  return (
+    <AppShell>
+      <AppShell.Header>
+        <AppShell.Title>Probe Hybridisation QC</AppShell.Title>
+      </AppShell.Header>
+      <AppShell.Main>
+        <Formik<ProbeHybridisationQCFormValues>
+          initialValues={formInitialValues}
+          validationSchema={validationWorkNumber}
```
seenanair, line 227: You are not validating time field which is required

src/pages/ProbeHybridisationQC.tsx
```diff
@@ -0,0 +1,391 @@
...
+                                  onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+                                    setFieldValue(`labwares.${labware.barcode}.completionDateTime`, e.target.value);
+                                  }}
+                                  value={values.labwares[labware.barcode]?.completionDateTime || ''}
+                                  className="w-1/2"
+                                  validate={validateCompletionDateTime}
+                                />
+                                <MutedText className="pl-2">
+                                  If not manually selected, the current timestamp will be applied automatically
+                                </MutedText>
```
seenanair, line 285: Please make use of formik error handling capabilities, by including this field in validation schema , so we can avoid using MutedText and validateCompletionDateTime

seenanair (commented):
Cool, looking good.
Just one question , the story says 'The labware must have had probe hybridisation already recorded on it.'. Of course the core  will return terror, but would it be better to let user know as soon as they scan labware?

src/pages/ProbeHybridisationQC.tsx
sabrine33: to cover the use case; where the user selects some options for a specific section and then adds other comments from the global comment

src/pages/ProbeHybridisationQC.tsx
sabrine33: the validation is done automatically by the input field, as I specified the min/max there

src/pages/ProbeHybridisationQC.tsx
sabrine33: hmmm I am not sure how to include this in the formik errors as the request is still valid, when the completion time is not been selected

src/pages/ProbeHybridisationQC.tsx
seenanair: Ohh, is it supposed to behave like that or is it to override everything by comments  from 'apply all' drop down when they change some value in it ? It is better to clarify with Katy.

src/pages/ProbeHybridisationQC.tsx
seenanair: I would think of it as having a default value which is the current time displayed in the field when it opens, and the validation can be based on that time which is the default

cypress/e2e/pages/ProbeHybridisationQC.cy.ts
```diff
@@ -0,0 +1,130 @@
...
+                          id: 10426
+                        }
+                      ]
+                    }
+                  })
+                );
+              }
+            )
+          );
+        });
```
seenanair, line 121: Please add a global mock handler for this as we do with all other operations in stan-client

src/components/Routes.tsx
```diff
@@ -200,6 +201,22 @@ export function Routes() {
...
         )}
       />
+      <AuthenticatedRoute
+        path="/lab/probe_hybridisation_qc"
+        render={(routeProps) => (
+          <DataFetcher
+            key={routeProps.location.key}
+            dataFetcher={() =>
+              stanCore.GetComments({
+                commentCategory: 'section',
```
seenanair, line 211: This should not be 'section', but Probe hybridisation comments I hope

seenanair (commented):
Story says , 'Probe hybridisation comments (editable on config page)' . i think this need to be added in config page

src/components/Routes.tsx
sabrine33: Good point, as the comments are applied for section, I thought we need to get the section comments?do  you think we need tp get the Probe QC comment instead ?

src/mocks/handlers/labwareHandlers.ts
```diff
@@ -115,27 +113,7 @@ const labwareHandlers = [
...
-          user: {
-            username: 'test',
-            role: UserRole.Normal
-          },
-          performed: ''
-        }
-      })
-    );
-  })
+  )
```
seenanair, line 116: I think this needs to be in this file

cypress/e2e/pages/ProbeHybridisationQC.cy.ts
```diff
@@ -67,9 +112,20 @@ describe('Probe Hybridisation QC', () => {
   });
 
   describe('Submission', () => {
-    context('when selecting a released labware', () => {
+    context('when  repeated comment specified for the same section', () => {
```
seenanair, line 115: The context name doesn't match with what it does

cypress/e2e/pages/xeniumProbeHybridisationSpec.cy.ts
```diff
@@ -266,7 +266,7 @@ describe('Xenium Probe Hybridisation', () => {
       });
 
       it('shows a success message', () => {
-        cy.findByText('Xenium probe hybridisation recorded on all labware').should('be.visible');
+        cy.findByText('Xenium probe hubridisation recorded on all labware').should('be.visible');
```
seenanair, line 269: Something gone wring with merge it seems. this typo was fixed in the other branch which is again reverted back now. Please change to 'hybridisation'

src/components/AppShell.tsx
```diff
@@ -251,19 +251,7 @@ function AppShell({ children }: AppShellParams) {
...
-                    caption={'Xenium'}
-                    icon={<LabwareIcon className="flex-shrink-0 h-6 w-6 text-sdb-400" />}
-                    description={'Recording Probe hybridisation'}
-                  >
-                    <NavLinkMenuItem
-                      caption={'Xenium Analyser'}
-                      path={'/lab/xenium_analyser'}
-                      icon={<LabwareIcon className="flex-shrink-0 h-6 w-6 text-sdb-400" />}
-                      description={'Recording Xenium analyser information'}
+                      description={'Recording Probe hybridisation for Xenium slides.'}
```
seenanair, line 254: Again merge - The previous description was correct 'Recording Xenium analyser information'

src/components/AppShell.tsx
```diff
@@ -458,9 +446,8 @@ function AppShell({ children }: AppShellParams) {
                     <div className="grid grid-cols-2 ml-2 gap-y-4 gap-x-8">
                       <StanMobileNavLink to="/lab/probe_hybridisation_xenium">Xenium</StanMobileNavLink>
                     </div>
-                    <h4 className="px-3 pt-2 text-sm font-normal ml-2 text-sp-600">Xenium</h4>
                     <div className="grid grid-cols-2 ml-2 gap-y-4 gap-x-8">
-                      <StanMobileNavLink to="/lab/xenium_analyser">Xenium Analyser</StanMobileNavLink>
```
seenanair, line 457: This shouldn't be removed

src/components/Table.tsx
```diff
@@ -105,13 +105,3 @@ export const TableCell: React.FC<
...
-
-type Props = {
-  children: string | ReactNode | ReactNode[];
-};
-export const TabelSubHeader: React.FC<Props> = ({ children, ...rest }) => {
-  return <div className="flex text-xs font-medium text-gray-500 uppercase">{children}</div>;
-};
-export const TabelCentredCell: React.FC<Props> = ({ children, ...rest }) => {
-  return <div className="flex items-center justify-center">{children}</div>;
-};
```
seenanair, line 117: This as well need to be kept

src/graphql/queries/FindLatestOperation.graphql
```diff
@@ -1,5 +1,5 @@
 query FindLatestOperation($barcode: String!, $operationType: String!) {
   findLatestOp(barcode: $barcode, operationType: $operationType) {
-    id
+    ...OperationFields
```
seenanair, line 3: This needs to be id not OperationFields

seenanair (commented):
It seems like  are some unintended changes happened with merge. Please check locally that everything works fine with your page and Xenium Analyser page

seenanair (approved):

PR #378, opened 2023-08-17
x1054 As ARA (Harriet) I would like the section position location (for multiple sections in a slot to appear on the labware details page and history page for ease of tracking #364

sabrine33:


src/components/history/History.tsx
```diff
@@ -96,7 +96,7 @@ export default function History(props: HistoryProps) {
         accessor: 'sectionNumber'
       },
       {
-        Header: 'Slot Id',
+        Header: 'Address',
         accessor: 'slotId'
```
seenanair, line 100: For 'Address' column , why the accessor is given as 'slotId' instead of address?

seenanair (approved):

PR #381, opened 2023-08-18
X1061_visium_cytassist_Select_box_defaults_to_blank

sabrine33:


cypress/e2e/pages/cytAssist.cy.ts
```diff
@@ -8,8 +8,8 @@ describe('CytAssist Page', () => {
   });
 
   describe('On load', () => {
-    it('shows a Visium LP CytAssist slide for the output', () => {
-      cy.findAllByText(/Visium LP CytAssist/i).should('have.length.above', 0);
+    it('output labware type should be empty', () => {
+      cy.findByTestId('output-labware-type').should('have.value', '');
```
seenanair, line 12: Would it better to use shouldDisplaySelectedValue and check against the value here?

cypress/e2e/pages/cytAssist.cy.ts
```diff
@@ -32,6 +32,15 @@ describe('CytAssist Page', () => {
     });
   });
 
+  context('When user selects Visium LP CytAssist labwareType', () => {
+    before(() => {
+      selectLabwareType(LabwareTypeName.VISIUM_LP_CYTASSIST);
+    });
+    it('shows a Visium LP CytAssist slide for the output', () => {
+      cy.findAllByText(/Visium LP CytAssist/i).should('have.length.above', 0);
```
seenanair, line 40: same as above - shouldDisplaySelectedValue

seenanair (commented):
I have found an issue with selection of labware type. Please test it

seenanair:
@sabrine33 It is not letting to select Visium LP CytAssist initially also not letting to change back to em[pty option

https://github.com/sanger/stan-client/assets/55585488/0329b10a-e650-46f0-b784-0ee76c7d0aed

seenanair (commented):
There is still a minor issue that the labware type cannot be resetted back to blank, once one of those labware type is selected.
I think we are not sending labware type as empty/undefined to the state machine
But, this can probably cause more work , so can probably leave it for the time being

seenanair (approved):

seenanair (approved):

PR #384, opened 2023-08-23
fixes RT#786191 Bug: client shows "uploading" and never completes#380

sabrine33:


src/components/upload/useUpload.ts
```diff
@@ -36,15 +36,22 @@ export function useUpload(url: string, errorField?: string) {
       postUpload(url, file)
         .then((response) => {
           const success = response.ok;
-          response.json().then((response) => {
+          if (response.bodyUsed === false) {
```
seenanair, line 39: Please correct if I am wrong
To my understanding , bodyUsed is relevant  only when you're reading the response body multiple times or mutliple parts within your own code. This will ensure that we are not accidentally trying to read the body multiple times which is not happening here. Also here we are using json() method and json/ text methods automatically manage the bodyUsed property for you. 
So checking that may not add value, instead the  call to  _setUploadSuccess(true);_  in else block  you added fixed the problem as the issue is caused by  the way promise response is consumed it seems.
So, this change you made can still fix this even without using bodyUsed , perhaps
For. eg
 **_postUpload(url, file)
        .then((response) => {
          if (response.ok) {
            setUploadSuccess(true);
          } else {
            // Consume the response body 
            return response.json();
          }
        })
        .then((data) => {
          setUploadResponse(data);
          setUploadSuccess(true);
        })
        .catch((error) => {
          setError(new Error(error));
        });_**

sabrine33: Hi @seenanair , yes, you are completely correct.  But the issue here is caused when invoking response.json() because the requests submitted from the file manager do not include a JSON response in their body. Therefore, the body is consumed within the first call of (then).

The solution above will work fine for the file manager, but it might break the workflow of other components (ie, section registration) as we still need to read the json response when it is ok

seenanair: Yes, file manger doesn't have json response. Sorry, I didn't mean that exact code, something to deal with both cases , to set setUploadSuccess(true); as soon as we know the response.ok is true, otherwise deal with the json response ?
postUpload(url, file)
        .then((response) => {
          const success = response.ok;
          setUploadSuccess(success);
          response.json().then((response) => {
            if (!success) {
              // get error message from body
              setError(new Error(errorField ? response[errorField] : response.message));
            } else {
              setUploadResponse(response);
            }
          });
        })
        .catch((error) => {
          // get error message from body or default to response status
          setError(new Error(error));
        });

src/components/upload/useUpload.ts
```diff
@@ -19,7 +19,7 @@ export function useUpload(url: string, errorField?: string) {
 
   /**External request for upload**/
   const requestUpload = React.useCallback(
-    (file: File) => {
+    (file: File, setUploadInProgress) => {
```
seenanair, line 22: Just a doubt - Why  setUploadInProgress need to be passed on as a parameter to the callback? Can we not handle it in FileUploader itself?

sabrine33: The useEffect hook in the FileUploader has multiple dependencies. I found it more efficient to update the status of uploadInProgress in the useUpload component as soon as the upload process finishes and validate by its status in the useEffect.

seenanair (approved):

PR #390, opened 2023-08-30
fixes x1066 As PO (Minal), I want horizontal scroll bars at the top of the tables on the dashboards, history page, search page and SGP management page for ease of use#389

sabrine33:


src/components/TopScrollingBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+      setWidth(newWidth);
+    }
+  }, [getChildWrapperWidth, width]);
+
+  useEffect(() => {
+    setInnerDivWidth();
+
+    const topScrollBarDiv = outerDivRef.current;
+    const tableWrapper: HTMLDivElement | null | undefined =
+      childrenWrapperRef.current?.querySelector('[datatype="table-wrapper"]');
```
seenanair, line 29: Can we use 'ref' instead of this as it seems to be more efficent and are aware of react componet life cycle whereas  querySelector is an old style which scans all dom to get hold of elements.  It is not that, using querySelector is not correct or break but just a suggestion to use something in react world while we are using the framework.

src/components/TopScrollingBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+    const onTableWrapperScroll = () => {
+      tableWrapper.scrollLeft = topScrollBarDiv.scrollLeft;
+    };
+    topScrollBarDiv.addEventListener('scroll', onTableWrapperScroll);
+    const onTopBarScroll = () => {
+      topScrollBarDiv.scrollLeft = tableWrapper.scrollLeft;
+    };
+    tableWrapper.addEventListener('scroll', onTopBarScroll);
+
+    window.addEventListener('resize', setInnerDivWidth);
```
seenanair, line 41: just to remember for future work,  addEventListener is not recommended  to  use in React, but to use React's built-in event handling system, which works more effectively with React's virtual DOM and component lifecycle.
https://legacy.reactjs.org/docs/handling-events.html#:~:text=React%20events%20do%20not%20work,element%20after%20it%20is%20created.

src/components/TopScrollingBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+    return () => {
+      tableWrapper.removeEventListener('scroll', onTopBarScroll);
+      topScrollBarDiv.removeEventListener('scroll', onTableWrapperScroll);
+      window.removeEventListener('resize', setInnerDivWidth);
+    };
+  }, [width, setWidth, children, getChildWrapperWidth, setInnerDivWidth]);
+
+  return (
+    <div>
+      <div ref={outerDivRef} style={{ overflowX: 'auto', overflowY: 'hidden' }}>
```
seenanair, line 52: Please use tailwind classes to make it uniform and easy to mange

src/components/TopScrollingBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <div style={{ paddingTop: '1px', width }}>&nbsp;</div>
```
seenanair, line 53: tailwind classes

src/components/TopScrollingBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+      </div>
+      <div ref={childrenWrapperRef} style={{ overflow: 'auto', overflowY: 'hidden' }}>
```
seenanair, line 55: tailwind classes

seenanair (commented):
It's a big improvement on usability 👍 
I have a few questions
1) I think it will still display two horizontal scrollbars , the table's own bottom scrollbar and this custom one.  If so, if the table has got only less records (less than a page's height) , would it be ok display two scrollbars on top and bottom?

Hi @seenanair ,
Yes, it will display two horizontal scrollbars , the table's own bottom scrollbar and a custom one at the top of table. This is more for the table width (not height) but yes if the table have few columns, less that the window width, no scroll bar should be visible.

seenanair:
> It's a big improvement on usability 👍 I have a few questions
> 
> 1. I think it will still display two horizontal scrollbars , the table's own bottom scrollbar and this custom one.  If so, if the table has got only less records (less than a page's height) , would it be ok display two scrollbars on top and bottom?
> 
> Hi @seenanair , Yes, it will display two horizontal scrollbars , the table's own bottom scrollbar and a custom one at the top of table. This is more for the table width (not height) but yes if the table have few columns, less that the window width, no scroll bar should be visible.

No, I mean the height. The whole purpose of this horizontal scrollbar at top is to allow user to scroll easily in horizontal direction when table width is greater than the window and the table height exceeds the viewport size. 
 So, in cases If the height of the table  is less so that it fits the page and width is more than the window width, there will two scrollbars displayed. That might look a bit strange, it is worth checking it out and try on pages with less records

seenanair (approved):

PR #393, opened 2023-09-06
x1032 Add user full name to Release Recipient

sabrine33:


src/components/entityManager/EntityManager.tsx
```diff
@@ -66,8 +73,31 @@ type EntityManagerProps<E> = {
...
+  /**
+   * Extra property of the entity to display in the table
+   */
+  extraDisplayColumnName?: ExtraEntityColumn<E>;
+};
+
+type BasicReleaseRecipient = {
+  userFullName: string;
+  username: string;
+  enabled: boolean;
```
seenanair, line 86: Why this need to be redefined as we have already a graphql fragment - ReleaseRecipientFieldsFragment available?

src/pages/Configuration.tsx
```diff
@@ -347,7 +347,7 @@ export default function Configuration({ configuration }: ConfigurationParams) {
           also available on the <StyledLink to={'/sgp'}>SGP Management</StyledLink> page as "Work Requester"
         </p>
         <EntityManager
-          initialEntities={configuration.releaseRecipients}
+          initialEntities={convertToBasicReleaseRecipient(configuration.releaseRecipients)}
```
seenanair, line 350: Why we need these conversions?

seenanair (commented):
I have seen the following issues 
- The layout of columns are not correct
- The recipient names are not saved or updated

seenanair:
Screenshot 2023-09-06 at 15 16 55
Screenshot 2023-09-06 at 15 16 34

src/components/entityManager/EntityManager.tsx
```diff
@@ -66,8 +73,31 @@ type EntityManagerProps<E> = {
...
+): BasicReleaseRecipient[] => {
+  if (!Array.isArray(recipients)) {
+    recipients = [recipients];
+  }
+  return recipients.map(({ userFullName, username, enabled }) => ({
+    userFullName: userFullName || '',
+    username,
+    enabled
+  }));
+};
```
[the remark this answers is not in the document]
sabrine33: As the userFullName field is optional, the conversion is only to set it to empty if undefined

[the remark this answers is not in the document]
seenanair: Ok, is it for displaying purpose? If so, is it not possible to check it on the fly and display an empty string?

src/pages/Configuration.tsx
```diff
@@ -347,33 +347,49 @@ export default function Configuration({ configuration }: ConfigurationParams) {
...
           extraDisplayColumnName={{
-            label: 'User Full Name',
-            value: 'userFullName',
-            onChange: (username, userFullName) => {
-              return stanCore
-                .UpdateReleaseRecipientFullName({ username, userFullName })
-                .then((res) => convertToBasicReleaseRecipient(res.updateReleaseRecipientFullName)[0]);
+            label: 'Full Name',
+            value: 'fullName',
+            extraFieldPlaceholder: 'Enter User Full Name',
```
seenanair, line 384: Probably you can change this as well to  'Enter full name' to match column header

src/components/entityManager/EntityManager.tsx
```diff
@@ -290,7 +272,7 @@ export default function EntityManager<E extends Record<string, EntityValueType>>
           <TableCell colSpan={2}>
             <Input
```
seenanair, line 273: Can we have the class style for the full name <Input> to be as no border  (or a very subtle one) as in normal condition and when we start editing (that means when it gets focus), the border can appear? This way the field will look  more similar to other, I think

seenanair (commented):
Looking good 👍

seenanair (approved):

PR #398, opened 2023-09-08
Probe qc completion time

sabrine33:


src/pages/ProbeHybridisationQC.tsx
```diff
@@ -232,9 +230,7 @@ export default function ProbeHybridisationQC({ comments }: SectionComments) {
   );
 
   const validateCompletionDateTime = (selectedTime: string) => {
-    return new Date(selectedTime) > new Date(currentDateTime)
-      ? 'Please select a time on or before current time'
-      : undefined;
+    return new Date(selectedTime) > new Date() ? 'Please select a time on or before current time' : undefined;
```
seenanair, line 233: I would suggest to change this to global formik validation so that you don't need to compare every time. This will work without writing our own logic since we are already using Yup. 

 completionDateTime: Yup.date()
      .max(new Date(), 'Please select a date and time on or before current time')
      .required('Completion Time is a required field')
      .label('Completion Time')

seenanair (approved):

PR #408, opened 2023-09-25
mover fileManager and unrelease cypress tests to unit tests

sabrine33:


src/mocks/server.ts
```diff
@@ -0,0 +1,4 @@
+import { setupServer } from 'msw/node';
+import { handlers } from './handlers';
+
+export const server = setupServer(...handlers);
```
seenanair, line 4: There is /src/mocks/mswSetup.ts file. Probably this can be included there instead of a new file.

tests/unit/generic/utilities.ts
```diff
@@ -1,3 +1,68 @@
...
+  await userEvent.type(input, '{arrowDown}');
+  const option = screen.getAllByText(optionValue)[0];
+  expect(option).toBeInTheDocument();
+  await userEvent.click(option);
+};
+
+export const selectFocusBlur = async (dataTestId = 'select-div') => {
+  const selectBox = within(screen.getByTestId(dataTestId)).getByRole('combobox', { hidden: true });
+  await userEvent.click(selectBox);
+  await userEvent.tab();
```
seenanair, line 53: Can we not use `blur` method itself?

tests/unit/pages/fileManager.spec.tsx
```diff
@@ -0,0 +1,425 @@
...
+  await waitFor(() => {
+    fireEvent.change(screen.getByTestId('file-input'), { target: { files: [file] } });
+  });
+};
+
+const uploadFile = async () => {
+  await selectFile();
+  expect(screen.getByTestId('upload-btn')).toBeEnabled();
+  await userEvent.click(screen.getByTestId('upload-btn'));
+};
```
seenanair, line 425: The test case  when visting the page as an end user is missing. It can either be added here or as a cypress e2e test in fileManger.cy.ts

seenanair (commented):
Great job 👍  This is a very good piece of work and a very good improvement on STAN's test suite. Hopefully this way, a lot of e2e test cases can be converted to unit tests.

jest.config.js
```diff
@@ -6,5 +6,6 @@ module.exports = {
   },
   testEnvironment: 'jest-environment-jsdom',
   testMatch: ['<rootDir>/tests/unit/**/*.spec.{js,jsx,ts,tsx}'],
-  setupFilesAfterEnv: ['<rootDir>/src/setupTests.ts']
+  setupFilesAfterEnv: ['<rootDir>/src/setupTests.ts'],
+  transformIgnorePatterns: ['node_modules/(?!(msw)/)']
```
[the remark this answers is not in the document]
sabrine33: Most of the time, you don't want to transform your node_modules because it can be slow and libraries should be shipped in a way that they are already ready for usage in most environments. Here, node_modules are ignored from transformations, the msw module should still be transformed.

src/lib/sdk.ts
```diff
@@ -3,7 +3,9 @@ import Cookies from 'js-cookie';
 import React from 'react';
 import { getSdk } from '../types/sdk';
 import { NavigateFunction } from 'react-router-dom';
-export const graphQLClient = new GraphQLClient('/graphql');
+
+const endpoint = process.env.NODE_ENV === 'test' ? 'http://localhost:3000/graphql' : '/graphql';
+export const graphQLClient = new GraphQLClient(endpoint);
```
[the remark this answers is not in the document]
sabrine33: This code configures a GraphQL client to connect to different endpoints based on whether the application is under testing or in another environment (like production or development).

seenanair (approved):

PR #414, opened 2023-09-27
X1083 add rr names to release page

sabrine33:


src/components/forms/index.tsx
```diff
@@ -31,6 +31,63 @@ export type OptionTemplate<L extends string, V extends string, LV = string | num
...
+ * Extends OptionTemplate to allow for additional properties.
+ */
+export type ExtendedOptionTemplate<L extends string, V extends string, LV = string | number, VV = string | number> = {
+  [key in L]: LV;
+} & {
+  [key in V]: VV;
+} & {
+  [key: string]: any;
+};
+
```
seenanair, line 44: It is better to define ExtendedOptionTemplate type as an extension on OptionTemplate by ```
export type ExtendedOptionTemplate<L extends string, V extends string, LV = string | number, VV = string | number> = OptionTemplate<L, V, LV, VV> & {
  [key: string]: any;
}; 
```
This way it ensures that ExtendedOptionTemplate will include all the properties defined in OptionTemplate in case OptionTemplate gets modified in future

seenanair (approved):

sabrine33:
Merged into devel

PR #415, opened 2023-09-28
X1084 add costing to xenium probe hybe

sabrine33:


src/components/probeHybridisation/ProbeAddPanel.tsx
```diff
@@ -145,7 +186,15 @@ const ProbeAddPanel = ({ probePanels }: ProbeLotAddPanelProps) => {
...
                   const updatedLabware: ProbeOperationLabware = {
                     ...lw,
-                    probes: [...lw.probes, { name: probeName, lot: probeLot, plex: Number(probePlex) }]
+                    probes: [
+                      ...lw.probes,
+                      {
+                        name: probeName,
+                        lot: probeLot,
+                        plex: Number(probePlex),
+                        costing: probeCosting === 'SGP' ? SlideCosting.Sgp : SlideCosting.Faculty
```
seenanair, line 195: As probeCosting can be empty as well, it is better to compare with correct value other than ternary operation.

src/lib/machines/probeHybridisation/probeHybridisationMachine.ts
```diff
@@ -181,7 +182,7 @@ export const probeHybridisationMachine = createMachine<ProbeHybridisationContext
           if (!ctx.probeLabware.some((plw) => plw.barcode === lw.barcode)) {
             ctx.probeLabware.push({
               barcode: lw.barcode,
-              probes: [{ name: '', lot: '', plex: -1 }],
+              probes: [{ name: '', lot: '', plex: -1, costing: SlideCosting.Sgp }],
```
seenanair, line 185: I think costing has to be an empty string as initial value

src/lib/machines/probeHybridisation/probeHybridisationMachine.ts
```diff
@@ -215,7 +216,7 @@ export const probeHybridisationMachine = createMachine<ProbeHybridisationContext
         if (e.type !== 'ADD_PROBE_LOT') return;
         const probeLabware = ctx.probeLabware.find((plw) => plw.barcode === e.barcode);
         if (!probeLabware) return;
-        probeLabware.probes.push({ name: '', lot: '', plex: -1 });
+        probeLabware.probes.push({ name: '', lot: '', plex: -1, costing: SlideCosting.Sgp });
```
seenanair, line 219: Same here as well, costing field should be empty initially

src/components/probeHybridisation/ProbeAddPanel.tsx
sabrine33: probeCosting is a required field can not be empty, validation will not pass until the user selects a costing

seenanair: Ok

src/lib/machines/probeHybridisation/probeHybridisationMachine.ts
sabrine33: This is due to the type restriction,  I can not set the costing the empty/null as it is of tyoe SlideCosting, on the ui when the user adds a new probe row, the select box is set to empty

src/lib/machines/probeHybridisation/probeHybridisationMachine.ts
sabrine33: This is due to the type restriction,  I can not set the costing the empty/null as it is of tyoe SlideCosting, on the ui when the user adds a new probe row, the select box is set to empty

src/lib/machines/probeHybridisation/probeHybridisationMachine.ts
seenanair: Why can't the type be defined as SlideCosting | string or SlideCosting|undefined?

seenanair (approved):

PR #416, opened 2023-09-28
Release file options in Release page 

seenanair:
Added release options (by default all will be selected )
Added a download button from release page (with release optios displayed in release page, I hope it mkae sense to have download button displayed in release page as well)
Renamed "Select release file option" to "Change release file options" which navigates to "/releaseOptions' page as before

### DIFF
--- a/src/pages/Release.tsx
+++ b/src/pages/Release.tsx
@@ -4,6 +4,7 @@ import { Form, Formik } from 'formik';
 import {
   GetReleaseInfoQuery,
   LabwareFieldsFragment,
+  ReleaseFileOptionFieldsFragment,
   ReleaseLabware,
   ReleaseLabwareMutation,
   ReleaseRequest
@@ -35,6 +36,8 @@ import DataTable from '../components/DataTable';
 import RemoveButton from '../components/buttons/RemoveButton';
 import EditIcon from '../components/icons/EditIcon';
 import { useNavigate } from 'react-router-dom';
+import { Input } from '../components/forms/Input';
+import DownloadIcon from '../components/icons/DownloadIcon';
 
 const validationSchema = Yup.object().shape({
   releaseLabware: Yup.array()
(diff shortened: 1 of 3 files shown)


src/pages/Release.tsx
```diff
@@ -428,6 +433,43 @@ function Release({ releaseInfo }: PageParams) {
...
+                        {releaseInfo.releaseColumnOptions.map((releaseOption) => (
+                          <div className="flex flex-row items-center gap-x-2" key={releaseOption.displayName}>
+                            <Input
+                              type="checkbox"
+                              data-testid={`${releaseOption.displayName}-checkbox`}
+                              className={'w-5 rounded'}
+                              checked={selectedReleaseColumns.some(
+                                (column) =>
+                                  column.displayName === releaseOption.displayName &&
+                                  column.queryParamName &&
```
khelwood, line 448: Is this expression correct? And what is it for? `column.displayName === releaseOption.displayName && column.queryParamName && releaseOption.queryParamName`?

seenanair: This is to update the checkbox status (checked or unchecked), by comparing two `ReleaseFileOptionFieldsFragment` objects - one in  model data associated with each checkbox with the one in `selectedReleaseColumns`. The comparison compares both `displayName` and `queryParamName` of each `ReleaseFileOptionFieldsFragment` to ensure it is the same object.
Hope that make sense

khelwood: Should it say `column.queryParamName === releaseOption.queryParamName` instead of `column.queryParamName && releaseOption.queryParamName`

seenanair: Ohh,  I didn't see it. You are absolutely right. I will change it now

seenanair: Done 👍

sabrine33 (approved):

PR #420, opened 2023-10-03
add extraction method to the extraction page

sabrine33:


src/lib/machines/extraction/extractionMachine.ts
```diff
@@ -123,12 +134,13 @@ export const extractionMachine = createMachine<ExtractionContext, ExtractionEven
...
             labwareType: 'Tube',
-            barcodes: ctx.labwares.map((lw) => lw.barcode)
+            barcodes: ctx.labwares.map((lw) => lw.barcode),
+            equipmentId: ctx.equipmentId > 0 ? ctx.equipmentId : undefined
           }
         });
       },
       validateExtraction: (ctx: ExtractionContext) => (send) => {
-        const isValid = ctx.labwares.length > 0 && ctx.workNumber !== '';
+        const isValid = ctx.labwares.length > 0 && ctx.workNumber !== '' && ctx.equipmentId !== 0;
```
seenanair, line 143: It would be better to use ctx.equipmentId>0, as in the request , it considers equipmentId > 0 as valid

src/mocks/repositories/releaseRecipientRepository.ts
```diff
@@ -12,7 +12,7 @@ const seeds: Array<ReleaseRecipientFieldsFragment> = [
   releaseRecipientFactory.build({ username: 'lh7', fullName: 'Liam Hickey' }),
   releaseRecipientFactory.build({ username: 'vk8' }),
   releaseRecipientFactory.build({ username: 'cc36' }),
-  releaseRecipientFactory.build({ username: 'aw24' }),
+  releaseRecipientFactory.build({ username: 'aw24', fullName: 'Andrew Wiles' }),
```
seenanair, line 15: What is this change about in this context?

src/pages/Extraction.tsx
```diff
@@ -46,13 +48,34 @@ function buildExtractionTableData(ctx: ExtractionContext) {
...
+  equipments: EquipmentFieldsFragment[];
+};
+
+const extractionEquipments = (equipments: EquipmentFieldsFragment[]) => {
+  const methods: Array<{ value: number; label: string }> = [];
+
+  equipments.forEach((equipment) => {
+    methods.push({ value: equipment.id, label: `Automated - ${equipment.name}` });
+  });
+  methods.push({ value: -1, label: 'Manual' });
```
seenanair, line 64: Why the value is -1? While sending request equipmentId<0 is considered as undefined

src/pages/Extraction.tsx
```diff
@@ -46,13 +48,34 @@ function buildExtractionTableData(ctx: ExtractionContext) {
...
+ */
+type ExtractionEquipments = {
+  equipments: EquipmentFieldsFragment[];
+};
+
+const extractionEquipments = (equipments: EquipmentFieldsFragment[]) => {
+  const methods: Array<{ value: number; label: string }> = [];
+
+  equipments.forEach((equipment) => {
+    methods.push({ value: equipment.id, label: `Automated - ${equipment.name}` });
```
seenanair, line 62: If not confirmed already, it is good to communicate to users that 'Automated"  is appended to all new equipment ids in config page. This helps to avoid 'Automated' as a part of equipment name when new ones are added by them in config page

seenanair (commented):
Looks good. Few things to mention
1) Story says, the automated Plus machines (edited on config). I couldn't find any changes to allow editing on config page
2) Commonly, even if we have unit test cases for all components, it is good to have few e2e test cases that just gives a normal health check for the page to see all components interact with each other as it is supposed to be.


_1) The extraction machines should be added in the equipment config page within a category ='extract'
2) Technically that it is very true, and it would be just easier to keep them/update them, but as they generate lots of false negative and generally quite slow, I thought the procedure that we agree for to try to migrate as much as possible, at least for the self contained component to unit tests and drop the e2e tests. Based on that, I thought this is component is a good example to do so._

src/mocks/repositories/releaseRecipientRepository.ts
sabrine33: The release tests keep randomly failing as it is not able to find the full user in the select list, as it is used to be in the bottom, I just used to a user name that can be at the top (as the list is alphabetic sorted) to stop the release tests from failing.

src/pages/Extraction.tsx
sabrine33: hmm, I think it is quite confusing to add an equipment with 'Automated' as prefix also, as that will not much the way the other equipment'category

seenanair: Probably, it can added as a note in config page, in equipments tab for this category that "Automated" will be prefixed , so that it will be conveyed properly

sabrine33: oh do mean we prefix the added equipments in the config page, I mean the user added a machine 'EZ2' and we stores as 'Automated - EZ2' ?

seenanair:
> Looks good. Few things to mention
> 
> 1. Story says, the automated Plus machines (edited on config). I couldn't find any changes to allow editing on config page
> 2. Commonly, even if we have unit test cases for all components, it is good to have few e2e test cases that just gives a normal health check for the page to see all components interact with each other as it is supposed to be.
> 
> _1) The extraction machines should be added in the equipment config page within a category ='extract' 2) Technically that it is very true, and it would be just easier to keep them/update them, but as they generate lots of false negative and generally quite slow, I thought the procedure that we agree for to try to migrate as much as possible, at least for the self contained component to unit tests and drop the e2e tests. Based on that, I thought this is component is a good example to do so._

1) Ok
2) I totally agree and what you have done adds a lot of value. This addition definitely aligns with what we have planned which is  to convert as many as possible e2e test cases to be part of unit testing, so that e2e test cases will be minimal.
It's up to the discretion of the tester as well as bit grey on what the boundaries are. But , what  I meant is as e2e actually puts in all components together, just  focusing on main operation ensures robustness  for e.g just one in this case, scan a labware-> sgp selection -> equipment selection -> click extract  and check the operation is completed and works normally.

src/pages/Extraction.tsx
seenanair: No, I meant just to give a textual indication to the user in config page that whatever machine they will be added will be displayed with "Automated" prefix in the page. It is, just a thought, but probably , it can be tested by user and see how they think about it.

seenanair: If the user thinks it's ok, then it's perfectly fine

src/pages/Extraction.tsx
sabrine33: the 'value' can only hold a number or string, as it is assigned to 'OptionTemplate' type. as the method is manual, no equipment id should be assigned.

seenanair: Ok, got it, that makes sense.  👍

seenanair (approved):

PR #425, opened 2023-10-09
migrate location cypress tests to jest tests

sabrine33:


seenanair (approved):

seenanair:
@sabrine33 Maybe it is better to move these e2e tests in jests to a different folder to make it more clear.

PR #427, opened 2023-10-10
X1081 update reset button for sectioning page

sabrine33:


tests/unit/components/sectioning/sectioningConfirmation.spec.tsx
```diff
@@ -0,0 +1,219 @@
...
+            })
+          };
+        });
+      });
+
+      it('reloads to Sectioning - Plan page when Reset Form button is clicked', () => {
+        waitFor(async () => {
+          await userEvent.click(screen.getByRole('button', { name: 'Reset Form' }));
+          expect(screen.getByRole('heading', { level: 1 })).toHaveTextContent('Sectioning - Plan');
+          expect(screen.getByRole('button', { name: 'Next' })).toBeDisabled();
```
seenanair, line 65: While looking for the slotMeasurement issue, I saw an article recommending the following way of testing results of an action as as the waitFor function deals with the situation  when you we to wait for an element to appear or change, or when you expect some action to occur after a certain period of time.

Do you think this is a better way?

For e.g
```
 it('reloads to Sectioning - Plan page when Reset Form button is clicked', async () => {
          await userEvent.click(screen.getByRole('button', { name: 'Reset Form' }));
          await waitFor(() => {
              expect(screen.getByRole('heading', { level: 1 })).toHaveTextContent('Sectioning - Plan');
              expect(screen.getByRole('button', { name: 'Next' })).toBeDisabled();
         })
})
```

sabrine33: @seenanair technically that is the correct way to use it, but I found some tests fails if the action and the checks are not in the same await block.

PR #433, opened 2023-10-12
x1088 add equipment to rn analysis

sabrine33:


tests/unit/components/sectioning/sectioningConfirmation.spec.tsx
```diff
@@ -14,6 +14,21 @@ afterEach(() => {
...
+      return (
+        <select data-testid="workNumber">
+          <option value="SGP1008">SGP1008</option>
+          <option value="SGP1009">SGP1009</option>
+        </select>
+      );
+    })
+  };
+});
+
```
seenanair, line 31: I hope these are refinements

seenanair (approved):
I think  you have added unit test cases for almost all components in Analysis page. Only thing missing is for 'analysisLabware.machine', which I hope can be added later.

PR #436, opened 2023-10-16
Fixed bug - removed labware included in release list

seenanair:
Changed Formik function used for removing items to 'setValues'


khelwood:
If this bug is live on production, please make a new branch from the latest production version [2.28.1](https://github.com/sanger/stan-client/tree/2.28.1) and write the fix on there so we can release it to production.

khelwood:
Replaced by https://github.com/sanger/stan-client/pull/437

PR #439, opened 2023-10-17
x1097 search multiple donors and multiple external id

sabrine33:


src/pages/Search.tsx
```diff
@@ -62,9 +62,9 @@ const validationSchema = Yup.object()
 const emptyFindRequest: FindRequest = {
   createdMin: '',
   createdMax: '',
-  donorName: '',
+  donorNames: [],
```
khelwood, line 65: If the donor name/external name field is left blank, will the fields in the FindRequest be _empty_ or _null_? Because those will have different effects.

sabrine33: if not been populated by the user, they will be empty, should they be null in that case ?

khelwood: Yes they should be null

seenanair (commented):
I think you need to split the donor and external names to an array as the core is expecting an array of strings (core branch (find_multi) . I would suggest the best place would be in `serachService` just before you call the core api.
Also please update graphql to align with core

seenanair (approved):

PR #444, opened 2023-10-17
x1089 As TS (Katy) I would like to add which xenium machine is runnin…

sabrine33:


src/pages/XeniumAnalyser.tsx
```diff
@@ -47,6 +49,7 @@ type LabwareSamples = {
 export type XeniumAnalyserFormValues = {
   lotNumberA: string;
   lotNumberB: string;
+  equipmentId: number | undefined;
```
seenanair, line 52: Just to note, if you want to make it in a more ts way  you can always use equipmentId?: number which is the same as number|undefined

src/pages/XeniumAnalyser.tsx
```diff
@@ -103,6 +108,7 @@ const XeniumAnalyser = () => {
       .max(new Date(), 'Please select a date and time on or before current time')
       .required('Time is a required field')
       .label('Time'),
+    equipmentId: Yup.number().required().label('Equipment').required('Equipment is a required field'),
```
seenanair, line 111: Why `equipmentId`  is optional in form data structure and a required field in validation?

tests/unit/pages/lab/xeniumAnalyser.spec.tsx
```diff
@@ -139,6 +144,21 @@ describe('Xenium analyser', () => {
           });
         });
       });
+      describe('Equipment', () => {
+        it('should display an error message on blur when no equipment is selected', () => {
+          waitFor(async () => {
+            await selectFocusBlur('equipmentId');
+            expect(screen.getByText('Equipment is a required field')).toBeVisible();
```
seenanair, line 151: Just to understand, is this a better way to use it or await method where you do an action in act, outside of waitFor?

src/pages/XeniumAnalyser.tsx
sabrine33: so we can init the form with empty equipmentId option

seenanair: Ok

tests/unit/pages/lab/xeniumAnalyser.spec.tsx
sabrine33: It depends on the use case. In this scenario, we are waiting for elements to become visible, which can be contingent on an asynchronous operation. We need to wait for the elements to appear before taking action and asserting the results. An alternative approach is to use findBy with a promise, like this:
 _screen.findByText('equipmentId').then((el) => {
            el.blur(); 
            expect(screen.getByText('Equipment is a required field')).toBeVisible();
          });_
But also, _el.blur();_  is not advisable in most cases.  To ensure our tests closely mimic user behavior, it's better to use userEvent to interact with the element, as it provides a more realistic simulation of user actions."

seenanair (approved):

PR #452, opened 2023-10-23
x1101 add fixative and medium to lhe labware details display

sabrine33:


src/lib/helpers.ts
```diff
@@ -391,3 +391,12 @@ export function generateRandomIntegerInRange(min: number, max: number) {
...
 }
+
+/**
+ * Checks whether the given labware has samples in any of its slots.
+ * @param labware - The labware to check for samples.
+ * @returns `true` if samples are found in any slot, `false` otherwise.
+ */
+export const hasSamples = (labware: LabwareFieldsFragment): boolean => {
+  return labware.slots.some((slot) => slot.samples && slot.samples.length > 0);
+};
```
seenanair, line 402: We are currently using helper.ts for generic function. The more appropriate place would be /src/lib/labwareHelper.ts

src/pages/LabwareDetails.tsx
```diff
@@ -89,6 +103,8 @@ export default function LabwareDetails() {
 
                     <StripyCardDetail term={'Date of Creation'}>{labware.created}</StripyCardDetail>
 
+                    {samplesDetails()}
```
seenanair, line 106: Just to note, may be for future, it is better to  avoid declaring a component within another component in same file. In this case , it is not that complex , but it can create performance issues in complex components.

seenanair (approved):

PR #458, opened 2023-10-27
x1092: Renaming

khelwood:
x1092: Rename in UI

Menu option "Sample processing" -> "Labware generation"
Mode "Block processing" -> "Block"
Mode "Pot processing" -> "Pot"

Operations and internals not renamed; I think they're fine as they are.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "start": "npm-run-all -p start:css start:js",
     "start:msw": "REACT_APP_MOCK_API=msw npm run start",
     "build": "npm-run-all build:css build:js",
-    "start:js": "react-scripts start",
+    "start:js": "react-scripts --openssl-legacy-provider start",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -60,7 +60,7 @@
     "start": "npm-run-all -p start:css start:js",
     "start:msw": "REACT_APP_MOCK_API=msw npm run start",
     "build": "npm-run-all build:css build:js",
-    "start:js": "react-scripts start",
+    "start:js": "react-scripts --openssl-legacy-provider start",
```
seenanair, line 63: Why this flag is required?

khelwood: The client tests won't run on my computer without the flag. If it causes problems, feel free to remove it.

seenanair: We can change it in case it create problems.

seenanair (approved):

PR #471, opened 2023-11-08
X1107 - Pre treatment qc in Stain QC page

seenanair:


cypress/e2e/pages/stainingQC.cy.ts
```diff
@@ -131,5 +131,42 @@ describe('Staining QC', () => {
...
+      });
+      it('should not display Pass/Fail icons', () => {
+        cy.findAllByTestId('passIcon').should('have.length', 0);
+      });
+      it('has comment dropdowns enabled', () => {
+        getAllSelect('comment').forEach((elem: any) => {
+          cy.wrap(elem).should('be.enabled');
+        });
+      });
+      context('when changing the comment all dropdown', () => {
```
sabrine33, line 154: will it better if  we move the page setting /loading checks  to unit tests ?

src/components/labwareResult/LabwareResult.tsx
```diff
@@ -180,7 +183,7 @@ export default function LabwareResult({
                   <FailIcon
```
sabrine33, line 183: is there a reason for replacing the non-null assertion by the optional chaining for the other fields here (other than result)

seenanair: Yes, result will become an optiona field on graphql request(core) with these new changes

seenanair: This is to avoid the situation where the component fails to load  or throws exception in case these fields are missing especially while calling from unit tests

cypress/e2e/pages/stainingQC.cy.ts
seenanair: Good point. But,  I have only added unit tests for LabwareResult, and  these tests need to be covered as part of StainingQC  unit tests which doesn't exist. I will see whether it can be added

sabrine33 (approved):

PR #475, opened 2023-11-16
fixes x1116 (client) As TS (Katy) I would like any user to be able to…

sabrine33:


seenanair (commented):
Just one thought/suggestion - Most in-place operations like Destroy,Release/Unrelease all have similar layouts. I assume this being a kind of similar operation , it would be good to have similar layout to improve consistency.

_@seenanair, a very valid suggestion, I tried to use the same layout in the first place, but as the suggested table is wider here (more columns requested to be displayed), I thought to take out the summary bit from the page, so we can use the most of the space to the table._

seenanair:
> Just one thought/suggestion - Most in-place operations like Destroy,Release/Unrelease all have similar layouts. I assume this being a kind of similar operation , it would be good to have similar layout to improve consistency.
> 
> _@seenanair, a very valid suggestion, I tried to use the same layout in the first place, but as the suggested table is wider here (more columns requested to be displayed), I thought to take out the summary bit from the page, so we can use the most of the space to the table._

@sabrine33 If it helps, I think for Release page, we are using a wider table but with a horizontal scrollbar.

seenanair (approved):

sabrine33:
merged into devel

PR #481, opened 2023-11-20
Bug fix

seenanair:
Bug fix for issue - xenium QC comments seem to be pooling from all the comments on the config page


khelwood (approved):
I am going to merge this into a new branch so we can add it to the current release candidate without adding anything else.

PR #494, opened 2023-11-29
X890 get replicate number from source

sabrine33:


src/components/originalSampleProcessing/blockProcessing/BlockProcessingLabwarePlan.tsx
```diff
@@ -105,6 +94,7 @@ const BlockProcessingLabwarePlan = React.forwardRef<HTMLDivElement, BlockProcess
     const { requestError, layoutPlan } = current.context;
     const { layoutMachine } = current.children;
     const { setFieldValue, values } = useFormikContext<BlockFormData>();
+    const [disableRepNumber, setDisableRepNumber] = useState<boolean>(false);
```
seenanair, line 97: We can use values.plans[rowIndex].replicateNumber instead of this which avoids managing another react state.

src/components/originalSampleProcessing/blockProcessing/BlockProcessingLabwarePlan.tsx
```diff
@@ -133,9 +119,11 @@ const BlockProcessingLabwarePlan = React.forwardRef<HTMLDivElement, BlockProcess
       const plannedActions: Source[] | undefined = layoutPlan.plannedActions.get('A1');
       if (plannedActions && plannedActions.length > 0) {
         setFieldValue(`plans.${rowIndex}.sourceBarcode`, plannedActions[0].labware.barcode);
+        setFieldValue(`plans.${rowIndex}.replicateNumber`, plannedActions[0].replicateNumber);
+        setDisableRepNumber(plannedActions[0].replicateNumber!.length > 0);
```
seenanair, line 123: As mentioned before, this is not required as we already setting plans.${rowIndex}.replicateNumber with the value .

src/components/originalSampleProcessing/blockProcessing/BlockProcessingLabwarePlan.tsx
```diff
@@ -176,7 +164,7 @@ const BlockProcessingLabwarePlan = React.forwardRef<HTMLDivElement, BlockProcess
                   name={`plans.${rowIndex}.replicateNumber`}
                   label={'Replicate Number'}
                   type={'text'}
-                  disabled={true}
+                  disabled={disableRepNumber}
```
seenanair, line 167: This can be replaced wiuth values.plans[rowIndex]?.replicateNumber

seenanair: Also, if you check in console this input is producing a react warning for changing from uncontrolled to controlled input which happens when an "undefined" value is passed to a component . This can be rectified by providing a value={values.plans[rowIndex] && values.plans[rowIndex].replicateNumber?values.plans[rowIndex].replicateNumber
                      : ''
                  }

src/components/originalSampleProcessing/blockProcessing/BlockProcessingLabwarePlan.tsx
sabrine33: using the  plans.${rowIndex}.replicateNumber check instead will disable the input field once the user enters a value (use case where the source does not have a replicate number, at soon as the user enters a value, the field disables)

seenanair: Ok,that's correct. So disabled status need to be solely based on given layoutPlan I suppose. In that case, please ignore all those comments related.

seenanair (approved):

sabrine33:
merged to devel

PR #500, opened 2023-12-05
destroyed labware fails on UAT

sabrine33:


src/pages/Reactivate.tsx
```diff
@@ -62,7 +62,7 @@ export const Reactivate = () => {
 
   const isLabwareInactive = useCallback(
     (labwares: LabwareFieldsFragment[], foundLabware: LabwareFieldsFragment): string[] => {
-      return foundLabware.discarded === false || !foundLabware.destroyed === false
+      return foundLabware.discarded === false && foundLabware.destroyed === false
```
seenanair, line 65: inactive doesn't mean either discarded or destroyed?

PR #503, opened 2023-12-07
provide the flag labware admin page and the update the labware details page to dislay the recorded flags

sabrine33:


src/pages/FlagLabware.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const checkRelatedFlagges = useCallback(
+    async (labwares: LabwareFieldsFragment[], foundLabware: LabwareFieldsFragment): Promise<string[]> => {
+      return stanCore
+        .GetLabwareFlagDetails({
+          barcodes: [foundLabware.barcode]
+        })
+        .then((response) => {
+          setRelatedFlags(response.labwareFlagDetails);
+          return [];
+        });
```
seenanair, line 81: Using await will improve readability, also allows  try/catch blocks for error handling

src/pages/FlagLabware.tsx
```diff
@@ -0,0 +1,183 @@
...
+    },
+    [stanCore]
```
seenanair, line 83: Would it not require `setRelatedFlags` in dependency array?

src/pages/FlagLabware.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+                              {relatedFlags.length > 0 && <LabwareFlagDetails flagDetails={relatedFlags} />}
+
+                              <div className="mt-6">
+                                <Heading level={3}>Description</Heading>
+                                <MutedText>Please enter a reason for flagging the labware.</MutedText>
+
+                                <Field
+                                  as={'textarea'}
+                                  placeholder="Enter your the reason for flagging the labware here."
```
seenanair, line 135: Is it required to have both MutedText and placeholder with the same information?

src/pages/FlagLabware.tsx
sabrine33: React guarantees that setState function identity is stable and won't change on re-renders. This is why it's safe to omit from the useEffect or useCallback dependency list.

src/pages/FlagLabware.tsx
sabrine33: not required, I will remove the placeholder and keep the muted text, so it is coherent with the other pages

seenanair (approved):

sabrine33:
merged to devel

PR #512, opened 2023-12-12
Bug fix for equipment dropdown missing

seenanair:
Replaced data fetcher component with React 18 loader/useLoader functionality which asynchronously fetches the equipments.

### DIFF
--- a/src/components/RouteLayout.tsx
+++ b/src/components/RouteLayout.tsx
@@ -322,23 +322,22 @@ const RouteLayout = () => {
         <Route element={<AuthLayout />}>
           <Route
             path="/lab/imaging"
+            loader={async () => {
+              const recordInPlaceInfo = await stanCore.GetRecordInPlaceInfo({ category: 'scanner' });
+              return recordInPlaceInfo.equipments;
+            }}
             element={
-              <DataFetcher dataFetcher={() => stanCore.GetRecordInPlaceInfo({ category: 'scanner' })}>
-                {(recordInPlaceInfo) => (
-                  <RecordInPlace
-                    title={'Imaging'}
-                    operationType={'image'}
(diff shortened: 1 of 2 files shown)


khelwood:
What caused the field to be missing?

seenanair:
@khelwood 

The page requires 'equipments' to be fetched from the core to populate the dropdown, and the dropdown visibility is based on whether there are equipments or not.

There are two fixes - 
1) I believe the `equipment` field should be visible in all cases, whether there are equipments or not. So, I have removed that checking so that even in cases where it couldn't fetch equipments, it will display an empty dropdown box, which I believe is more intuitive.

2) Currently, the fetching of equipments is achieved using a higher-order component (HOC) called `DataFetcher`, which fetches data and passes it as props to the component. However, with the React 18 upgrade, a new approach has been introduced using the loader and the useLoader hook. This facilitates asynchronous calls to fetch any props necessary for page loading.
Despite the React 18 upgrade, the existing approach has been retained as it was working and to avoid extensive changes, given its widespread use across many pages. 
But as it is showing inconsistencies, this is now changed to loader function.

Also, I think it would be better to replace all the existing pages using this loader function, as this is the second time a similar issue has been reported. Therefore, I have added a story for it.
https://github.com/orgs/sanger/projects/9/views/2?pane=issue&itemId=47310511

khelwood:
Sounds good. If this is a change to the `recordInPlace` page, have you checked there are no uses of that page that are not linked to an equipment?

seenanair:
> Sounds good. If this is a change to the `recordInPlace` page, have you checked there are no uses of that page that are not linked to an equipment?

Updated

khelwood (approved):

PR #516, opened 2023-12-12
update reactivate to read from work status comments

sabrine33:


sabrine33:
@seenanair  this is the same "reactivate" branch checkout from the last tagged branch

src/components/workAllocation/WorkAllocation.tsx
```diff
@@ -382,25 +382,28 @@ export default function WorkAllocation() {
...
                         .then((study) => {
                           if (study && study.dnapStudy) {
                             setFieldValue('studyName', study.dnapStudy.name);
-                            setFieldValue('ssStudyId', ssId);
+                          } else {
+                            setFieldValue('studyName', 'undefined');
                           }
                         })
                         .catch((e) => {
                           setFieldValue('studyName', 'undefined');
```
seenanair, line 398: Why we need to set this to 'undefined' string?. Would it be better to make it to an empty string?

sabrine33: it seems like when setting to an empty string, when validate to read it fails  (ie: values.studyName return false )

seenanair: Where are we validating this?

seenanair: I think in the default form value as well it can be initialised to '' instead of undefined

sabrine33: The validation is done just below L403, to display the study name of an error message

seenanair: I think it shouldn't be a problem because TS validates an empty string as false as well when used with && or | operator -  “falsy” values (e.g. the empty string, 0, NaN, and, well, false)
(https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html)

sabrine33: that's exactly the issue, as I only need to display the error message when the user enters invalid study id

seenanair (approved):

PR #536, opened 2024-01-22
adding a new operation visium QC operation : qPcr results

sabrine33:


src/components/visiumQC/QPcrResults.tsx
```diff
@@ -0,0 +1,151 @@
...
+  );
+
+  const handleChangeAllMeasurements = React.useCallback(
+    (measurementName: string, measurementValue: string) => {
+      //Reset Errors
+      setErrors({});
+      setTouched({});
+      const measurements = values?.slotMeasurements ? [...values.slotMeasurements] : [];
+      measurements
+        ?.filter((measurement) => measurement.name === measurementName)
```
seenanair, line 64: Why we need to filter it? I thougt we only have one measurement (Cq value) here? or Is it for future additions?

src/pages/VisiumQC.tsx
```diff
@@ -180,14 +182,15 @@ export default function VisiumQC() {
...
+      values.qcType === QCType.QPCR_RESULTS ||
       (values.qcType === QCType.CDNA_AMPLIFICATION && values.slotMeasurements)
     ) {
       sendCDNA({
         type: 'SUBMIT_FORM',
         values: {
           workNumber: values.workNumber,
           barcode: values.barcode,
-          slotMeasurements: values.slotMeasurements ?? [],
+          slotMeasurements: values.slotMeasurements?.map(({ samples, ...rest }) => rest) ?? [],
```
seenanair, line 193: Are we getting rid of 'samples' field? if so, why?

cypress/e2e/pages/visiumQC.cy.ts
```diff
@@ -574,6 +574,44 @@ describe('Visium QC Page', () => {
...
+
+        it('shows a success message', () => {
+          cy.findByText('qPCR results complete').should('be.visible');
+        });
+        after(() => {
+          cy.findByRole('button', { name: /Reset/i }).click();
+        });
+      });
+    });
+  });
```
seenanair, line 613: Couldn't find a test case when there is server error.

src/pages/VisiumQC.tsx
sabrine33: yes, because we don't need in the request, we are just adding some info in the display. also the GraphQL request will fail if we pass the wrong arguments .

seenanair (approved):

sabrine33:
merged to devel

PR #553, opened 2024-02-05
Self registration as end user

sabrine33:


tests/unit/pages/login.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+      act(() => {
+        fireEvent.click(screen.getByRole('button', { name: 'Register' }));
+      });
+      await waitFor(async () => {
+        expect(screen.getByText('Username is required')).toBeVisible();
+        expect(screen.getByText('Password is required')).toBeVisible();
+      });
+    });
+  });
+});
```
seenanair, line 37: It would be better to add a similar test case for 'Sign In' option as well.

src/pages/Login.tsx
```diff
@@ -134,17 +156,29 @@ const Login = (): JSX.Element => {
...
                       placeholder="Password"
                     />
                   </div>
 
                   <div className="mt-6">
-                    <LoginButton loading={formik.isSubmitting}>Sign In</LoginButton>
+                    <LoginButton loading={formik.isSubmitting} data-testid="signIn" disabled={formik.isSubmitting}>
+                      Sign In
+                    </LoginButton>
+                  </div>
```
seenanair, line 172: Just a UI thought, Would it be better to add an 'Or' in between Register and Sign In and to mention in  smaller font/Italics for 'Existing users' near to Sign In ?

tests/unit/pages/login.spec.tsx
seenanair: Also, just a doubt , may be a core side question, @khelwood -  what will happen if an existing user clicks register button? Would it re-register an existing user or give a warning that user exists already?

seenanair (commented):
Looking good. 👍 Just commented a few minor things.

tests/unit/pages/login.spec.tsx
sabrine33: @seenanair,in case a user registers with existing credentials, the core handles it as a standard login process

seenanair (approved):

PR #554, opened 2024-02-06
Displaying flagged labware section in history page

seenanair:
if there are any flags in the history, it displays a box above the history table listing the flagged labware.

### DIFF
--- a/cypress/e2e/pages/history.cy.ts
+++ b/cypress/e2e/pages/history.cy.ts
@@ -216,38 +216,6 @@ describe('History Page', () => {
       });
     });
   });
-
-  describe('when history includes a flagged labware', () => {
-    before(() => {
-      cy.msw().then(({ worker, graphql }) => {
-        const historyEntry = buildHistory('SGP1008', true);
-        debugger;
-        worker.use(
-          graphql.query<FindHistoryQuery, FindHistoryQueryVariables>('FindHistory', () => {
-            return HttpResponse.json({
-              data: {
-                history: historyEntry
-              },
-              once: true
-            });
-          })
-        );
-      });
-      cy.visitAsAdmin('/history?workNumber=SGP1008');
-    });
-    it('displays a Flagged labware section', () => {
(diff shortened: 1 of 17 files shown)


cypress/e2e/pages/history.cy.ts
```diff
@@ -214,6 +217,37 @@ describe('History Page', () => {
...
+      cy.findByText('Flagged Labware').should('exist');
+      cy.findByTestId('flagged-labware').should('exist');
+    });
+    it('should navigate to  labware page when clicked', () => {
+      cy.findByTestId('flagged-labware').within(() => {
+        cy.findByText('STAN-1000').should('exist');
+        cy.findByText('STAN-1000').click();
+      });
+      cy.url().should('be.equal', 'http://localhost:3000/labware/STAN-1000');
+    });
```
sabrine33, line 248: Should these test cases be part of the unit tests, as they don't require any interaction with other components

src/components/history/History.tsx
```diff
@@ -205,6 +212,27 @@ export default function History(props: HistoryUrlParams) {
...
+          .filter((labware) => labware.labwareFlagged.flagged)
+          .map((labware) => labware.labwareFlagged);
+        setFlaggedLabware(labwareFlagged);
+      } catch (error) {
+        setFlaggedLabware([]);
+      }
+    }
+    flaggedLabwareFromHistory(history);
+  }, [history, setFlaggedLabware]);
+
```
sabrine33, line 235: I am not sure, but I thought we can check if the labware is flagged from the history search result already, no need to send another request to the core, like checking by event type (if contains 'Flag labware')

seenanair: I don't think it is possible l by looking for "Flag labware" events in the history, because the labware might inherit a flag even if the flag event isn't included in the history you're looking at. David will be adding this information in the history query , so the client can be refactored after this.

sabrine33 (approved):

PR #558, opened 2024-02-14
Bug fix for x1143

seenanair:
- Display of  Flagged labware section  for events that is not associated with SGPs
- Removed Flagged labware section from labware details page

### DIFF
--- a/src/components/history/History.tsx
+++ b/src/components/history/History.tsx
@@ -25,14 +25,14 @@ export default function History(props: HistoryProps) {
   const historyMachine = React.useMemo(() => {
     return createHistoryMachine();
   }, []);
-  const getHistoryURLParmas = (props: HistoryProps): HistoryUrlParams => {
+  const getHistoryURLParams = (props: HistoryProps): HistoryUrlParams => {
     const { displayFlaggedLabware, ...urlProps } = props;
     return urlProps;
   };
   const { displayFlaggedLabware, ...urlProps } = props;
   const [current, send] = useMachine(historyMachine, {
     context: {
-      historyProps: getHistoryURLParmas(urlProps)
+      historyProps: getHistoryURLParams(urlProps)
     }
   });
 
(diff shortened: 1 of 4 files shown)


sabrine33 (approved):

src/components/history/History.tsx
```diff
@@ -20,13 +20,19 @@ import { HistoryUrlParams } from '../../pages/History';
 /**
  * Component for looking up and displaying the history of labware and samples
  */
-export default function History(props: HistoryUrlParams) {
+type HistoryProps = HistoryUrlParams & { displayFlaggedLabware?: boolean };
+export default function History(props: HistoryProps) {
   const historyMachine = React.useMemo(() => {
     return createHistoryMachine();
   }, []);
+  const getHistoryURLPramas = (props: HistoryProps): HistoryUrlParams => {
```
khelwood, line 28: `Pramas` -> `Params`

khelwood (commented):
Which bit of this solves the problem where the flagged-box was not showing up when it should?

seenanair:
> Which bit of this solves the problem where the flagged-box was not showing up when it should?

The issue was that the code, which displays the flagged labware section (lines 263 to 289), was within the condition uniqueWorkNumbers > 0. This condition would fail for events like 'Flag labware,' which is not associated with any SGP's. Therefore, it has been taken out of that condition. I hope this makes sense.

src/components/history/History.tsx
seenanair: Fixed

src/components/history/History.tsx
```diff
@@ -20,13 +20,19 @@ import { HistoryUrlParams } from '../../pages/History';
 /**
  * Component for looking up and displaying the history of labware and samples
  */
-export default function History(props: HistoryUrlParams) {
+type HistoryProps = HistoryUrlParams & { displayFlaggedLabware?: boolean };
+export default function History(props: HistoryProps) {
   const historyMachine = React.useMemo(() => {
     return createHistoryMachine();
   }, []);
+  const getHistoryURLParmas = (props: HistoryProps): HistoryUrlParams => {
```
khelwood, line 28: `Parmas` -> `Params`

khelwood (approved):

PR #559, opened 2024-02-14
Fixes for ESLint warnings and Storybook deployment error

seenanair:
Eslint warnings removed
Storybook build update

### DIFF
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -5,7 +5,7 @@ name: Deploy Storybook
 
 on:
   push:
-    branches: [fixes_eslint_warning_and_story_book_deployment]
+    branches: [devel]
 
 jobs:
   build:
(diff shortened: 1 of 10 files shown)


khelwood (commented):
Where there are changes like `let { container } =` and `jest.fn()`, what is the effect of those changes?

seenanair:
> Where there are changes like `let { container } =` and `jest.fn()`, what is the effect of those changes?

All these changes are to address the ESLint warnings

1) Eslint was giving a warning about  the unused variable  `container`  which is de-structured from the return value of the render method. As it was not used anywhere, it is removed to fix this warning
 ```
let { container } = 
        render(
          <LabwareResult
            labware={labware}
            initialLabwareResult={initialLwResult}/>
```
2) Object.assign(navigator, {
      clipboard: {
        writeText: () => {
          throw 'error';
        }
      }
    });
Here we are mocking a writeText method of clipboard, but the return value was not correct.
This change uses jest.fn() to create a mock function for the writeText method as we are only checking in our tests that whether this function is called or not and not the return value.

Hope it make sense.

PR #567, opened 2024-03-14
History graph

sabrine33:


khelwood (commented):
Is there a reason that creating a png file is preferable to saving the SVG data directly to an html file?

khelwood (approved):

sabrine33:
merged to devel

PR #570, opened 2024-03-14
for End Users, include their user names within the "work requester" select options

sabrine33:


khelwood (commented):
_My reading of this is that the client adds the current user to the recipients input field, so I should update the core to accept the current user as a special case in the recipient part of the request?_

I haven't reviewed how the core functionality works, but if it prevents the 'workRequester' from being the same as the current user, then yes, you should remove that restriction.

khelwood:
@sabrine33 It doesn't specifically prevent it; it just doesn't specifically allow it. The list of possible work requesters is a list that the client gets from the core. If the client specifies a requester that isn't on that list, the core will not accept it, unless I change the core to allow it.

khelwood (approved):

sabrine33:
merged to the devel

PR #575, opened 2024-03-14
Add N/A option to a life stage

sabrine33:


src/pages/registration/SectionForm.tsx
```diff
@@ -48,6 +48,16 @@ type SectionFormParams = {
...
 
+export const LifeStageMap: Record<string, string> = {
+  ...Object.entries(LifeStage).reduce(
+    (acc, [key, value]) => {
+      acc[key] = value;
+      return acc;
+    },
+    {} as Record<string, string>
+  ),
+  'N/A': 'n/a'
```
khelwood, line 59: What do the two versions of N/A represent here?

sabrine33: n/a for not applicable, I suppose, that's the ticket's request

khelwood: Yes, why are there two versions 'N/A' and 'n/a'? What do the two different versions represent?

sabrine33: oh, as the uppercase is for label, and the lowercase is for key, just to match how the other life stage values are set.

khelwood (commented):
The core expects null rather than `'n/a'` as a life stage value. Is there somewhere in here that translates `'n/a'` to null?

src/pages/registration/SectionForm.tsx
sabrine33: I checked it with the core and it looks like LifeStage is a required field. I am updating the branch, to send null to the core, but probably you need to do some work at your end.

khelwood: Yep there is a pull request in core to make LifeStage nullable. 👍

khelwood (approved):

sabrine33:
merged to devel

PR #580, opened 2024-03-19
apply the same or different SGP numbers to each labware scanned in for the Probe hybridisation QC

sabrine33:


khelwood (approved):

sabrine33:
merged to devel

PR #586, opened 2024-03-22
enhance library generation

sabrine33:
- stops PromptOnLeave from poping up after the request is sent
- Add Print label on request success


khelwood (approved):

sabrine33:
merged to devel

PR #594, opened 2024-03-26
remove query-string module and provide alternative methods

sabrine33:


src/lib/helpers.ts
```diff
@@ -102,20 +96,72 @@ export function safeParseQueryString<T>(params: SafeParseQueryStringParams<T>):
...
+    }
+  }
+
+  return parsedQuery;
+}
 
 /**
+ * @returns A function that takes a key and returns a function that encodes the key-value pair.
+ */
+function encoderForArrayFormat() {
```
khelwood, line 125: What are the things here being joined with the `join('')`?

khelwood (commented):
What prompted this change?

src/lib/helpers.ts
sabrine33: This function is primarily utilised for conducting searches with multiple values, such as on the work summary page, where users can search by multiple status values. Here, it concatenates the encoded key and value pair with the sign ([]=) and then encodes the resulting string. Finally, the entire array of key-value pairs is concatenated using the join method, which results in a single string containing all the encoded key-value pairs separated by the ampersand delimiter.

khelwood: @sabrine33 OK. Where you have `[encodeURIComponent(key), '[]'].join('')`, that is another way of writing `encodeURIComponent(key) + '[]'` I think.

khelwood: @sabrine33 OK. Where you have `[encodeURIComponent(key), '[]'].join('')`, that is another way of writing `encodeURIComponent(key) + '[]'` I think.

sabrine33: yes, basically we are joining the elements of array using the specified join ('' here )

khelwood: @sabrine33 What prompted the removal of the query-string module?

khelwood (approved):

sabrine33:
merged to devel

PR #609, opened 2024-04-02
add spatial location name to sectioning planning

sabrine33:
add spatial location name to sectioning planning and original sample registration


khelwood (approved):

sabrine33:
merged to devel

PR #618, opened 2024-04-04
Fix xenium probe e2e test

sabrine33:
xeniumProbeHybridisation cypress tests are randomly failing on the CI, with the error: element is not visible because its content is being clipped by one of its parent


khelwood (approved):

sabrine33:
merged to devel

PR #627, opened 2024-04-12
Cell Segmentation operation

sabrine33:


src/components/CellSegmentation/CellSegmentation.tsx
```diff
@@ -0,0 +1,263 @@
...
+      <LabwareScanner
+        onAdd={async (labware) => {
+          await setValues((prev: CellSegmentationFormProps) => {
+            let cellSegmentation = [...prev.cellSegmentation];
+            cellSegmentation.push({
+              labware: labware,
+              workNumber: prev.workNumberAll ?? '',
+              performed: prev.performedAll ?? '',
+              comments: prev.commentsAll ?? [],
+              costing: !isQc ? prev.costingAll ?? '' : undefined
```
khelwood, line 59: For `costing` here there are two defaults: `''` and `undefined`. Do they have separate use cases inside this field?

sabrine33: yes for QC, costing is not part of the form, so it should be set to undefined, for cell segmentation form, costing should be init to an empty string (as it is mandatory field there)

khelwood (approved):

sabrine33:
merged to devel

PR #635, opened 2024-04-18
RT-801468

sabrine33:


khelwood (approved):

khelwood:
@sabrine33 Is this a live bug that we want to release a fix for?

sabrine33:
> @sabrine33 Is this a live bug that we want to release a fix for?

@khelwood I believe so, as it is from RT

khelwood:
I've created an new rc with the fix, and merged that into devel

PR #655, opened 2024-04-26
add clean out operation

sabrine33:


cypress/e2e/pages/dualIndexPlate.cy.ts
```diff
@@ -230,16 +230,17 @@ describe('Dual Index Plate', () => {
...
+        selectOption('plateType', 'Fresh frozen - Dual Index TT Set A');
         cy.get('#sourceLabwares').within(() => {
           cy.findByText('A1').click();
         });
         cy.get('#destLabwares').within(() => {
           cy.findByText('A1').click();
         });
-        selectSGPNumber('SGP1008');
-        selectOption('plateType', 'Fresh frozen - Dual Index TT Set A');
+        cy.screenshot();
```
khelwood, line 243: Why has this `screenshot` been added?

sabrine33: The Cypress tests were failing on CI at that step, and I'm not sure why. I added the screenshot to check the reason, but then the CI passed. The instruction needs to be removed, though.

khelwood (approved):

sabrine33:
merged into devel

PR #663, opened 2024-05-01
download and redirect

sabrine33:
update release file link in the history page to redirect the user to release options page and download the file


khelwood (changes requested):
My reading of this is that it downloads the release-file and then redirects to the release-options page. I think it should just take you to the release-options page.

sabrine33:
@khelwood I thought that is what is been requested, may be worth double check with @KatyTaylor

khelwood (approved):

sabrine33:
merged into devel

PR #678, opened 2024-05-16
x1189 upload  xenium metrics summary to Stan

sabrine33:


cypress/e2e/pages/cytAssist.cy.ts
```diff
@@ -342,16 +342,15 @@ describe('CytAssist Page', () => {
             selectSlideCostings('Faculty');
             enterLOTNumber();
             enterProbeLOTNumber();
-            cy.screenshot();
             cy.findByTestId('external-barcode').within(() => {
-              cy.findByRole('textbox').clear().type('H1-9D8VN2V{enter}').blur();
```
khelwood, line 347: Are all these waits necessary for the tests to pass, or were they put in for debugging?

src/components/xeniumMetrics/MetricsReader.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+interface FileParserProps {
+  rowIndex: number;
+}
+
+/**
+ * Parses CSV data into an array of objects where each object represents a row in the CSV file.
+ * The keys of each object are column names (extracted from the CSV header), and the values
+ * are cell values from the corresponding row.
+ * are cell values from the corresponding row.
```
khelwood, line 20: This line is a duplicate of the line above

src/components/xeniumMetrics/MetricsReader.tsx
```diff
@@ -0,0 +1,256 @@
...
+ *
+ * @param csvData The CSV data to parse as a string.
+ * @returns An array of objects representing rows in the CSV file, with keys as column names
+ * and values as cell values.
+ */
+const parseCSV = (csvData: string): { [key: string]: string }[] => {
```
khelwood, line 26: Is there not an existing JS library that can handle parsing CSV? (and might also correctly check for quoted values)

cypress/e2e/pages/cytAssist.cy.ts
sabrine33: adding the waits stops the cytAssist tests from randomly failing. Adding the screenshot for debugging purposes  makes the tests pass, as the system pauses for sometime to take the screenshot.

src/components/xeniumMetrics/MetricsReader.tsx
sabrine33: nothing that I am aware off, quoted values should be handled okay ( the quotes are considered as part of the value)

khelwood: What I mean by "correctly check for quoted values" is that separator characters inside a quoted value are typically not interpreted as separators. So the line
```
Alpha, Beta, "Gam,ma"
```
would be parsed as three values: `Alpha`, `Beta` and `Gam,ma`.
It could be that this will never come up, and the values in metrics will never contain quotes or commas; but an existing JS library might just handle these things correctly so we wouldn't have to worry about it.

khelwood (approved):
Thanks for making that change.

sabrine33:
merged to devel

PR #694, opened 2024-06-06
new labware type with name 8 strip tube: 8 rows, 1 column

sabrine33:


khelwood (changes requested):
It looks fine but maybe you want to use the word `STRIP` in code instead of `STRIPE`, which has a different meaning.

khelwood (approved):

PR #699, opened 2024-06-17
fixing typeError in file manager

sabrine33:


khelwood (approved):
What caused the problem?

PR #706, opened 2024-06-21
Rt 803875 updating section number after deletion

sabrine33:
updating the section numbers after section/labware deletion when auto mode is on


khelwood (approved):

sabrine33:
merged into devel

PR #718, opened 2024-06-27
add extra details to the xenium analyser labware details table

sabrine33:
add the SGP number associated with the labware, probes used and the cell segmentation record on a Xenium labware


khelwood (approved):

sabrine33:
merged into devel

PR #720, opened 2024-07-01
x1213 add new omero project from sgp management page

sabrine33:


khelwood (commented):
I think _add omero project_ can only be performed by _admin_ users. I guess the option to add should only be visible if the user is admin, unless we're also changing the required user tier.

----
I'll update core to support endusers adding omero projects.

khelwood:
@KatyTudor Currently only admin users can add a new omero project. Do you want regular users and/or endusers to be able to add them?

KatyTudor:
@khelwood yes please, like the project and cost code on the SGP management page

khelwood (approved):

sabrine33:
merged into devel

PR #726, opened 2024-07-09
record the main peak size in addition to the concentration on the Vis…

sabrine33:


khelwood (approved):
Out of interest, why have all these test IDs been put in upper case?

sabrine33:
merged into devel

PR #729, opened 2024-07-15
DV200 value are not updated correct after updating the measurement value

sabrine33:


khelwood (commented):
Since this is a live bug, can you please create a branch from 2.44.3 and make the change on there?

PR #731, opened 2024-07-16
Rt 806566 transfer operation

sabrine33:
- Populate the source labware with its samples correctly, even before selecting a destination labware.
- Update the 'Output Labware Selection' checkbox accordingly to match the displayed labware in cases where there are multiple destination labware and the user is paginating through them.
- Fix a Visual bug in the Library generation operation, where when the user selects a different destination labware type, the pagination footer gets updated (i.e 2 of 2) even though only one destination labware is taken on consideration.

### DIFF
--- a/src/components/libraryGeneration/SlotCopyComponent.tsx
+++ b/src/components/libraryGeneration/SlotCopyComponent.tsx
@@ -132,7 +132,7 @@ export const SlotCopyDestinationConfigPanel: React.FC<DestinationLabwareScanPane
               locked={labware?.labware.barcode !== undefined}
               checkForCleanedOutAddresses
             >
-              {({ labwares, removeLabware, cleanedOutAddresses }) => {
+              {({ labwares, removeLabware }) => {
                 return (
                   <div className="flex flex-col">
                     {labwares.length > 0 && (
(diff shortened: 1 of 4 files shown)


src/components/libraryGeneration/SlotCopyComponent.tsx
```diff
@@ -381,6 +382,22 @@ function SlotCopyComponent({
...
+  const updateDestinationModeWhenPaginating = () => {
+    //Before selecting any destination
+    if (!destinationSelectionMode.current) {
+      return undefined;
+    }
+    if (destinationSelectionMode.current === DestinationSelectionMode.SCAN) {
+      return destinationSelectionMode.current;
+    }
+    if (selectedDestination.labwareType.name === '96 well plate') {
+      destinationSelectionMode.current = DestinationSelectionMode.PLATE;
```
khelwood, line 394: Why does `destinationSelectionMode.current` need to be set inside the function? Isn't the return value of this function sufficient?

sabrine33: destinationSelectionMode is a reference to the destination mode check list, here we are updating the destination mode accordingly to the labware type displayed.

khelwood (approved):

PR #771, opened 2024-10-04
add LP number select box to the CytAssist page

sabrine33:


khelwood (changes requested):
Unless I'm reading the code wrong, the page is only offering LP1 and LP20, and not offering the numbers in between.

khelwood (approved):

PR #777, opened 2024-10-15
X1244 number stored items

sabrine33:


src/components/RouteLayout.tsx
```diff
@@ -487,12 +488,22 @@ const RouteLayout = () => {
...
             if (params.locationBarcode) {
-              const res = await stanCore.FindLocationByBarcode({
-                barcode: params.locationBarcode
-              });
-              return res.location;
+              await stanCore
+                .FindLocationByBarcode({
+                  barcode: params.locationBarcode
+                })
+                .then(async (parent) => {
```
khelwood, line 497: It looks like the client is doing an extra lookup for each location contained in the current page's location. Have I got that right? and if so, why is that necessary?

sabrine33: as the information needs to be displayed on the parent leaf node, so I need to check if the children are leafs or not

khelwood: That makes sense. But the reason I added `numStored` to the graphql was so that the client would not have to load every child location. If you have to anyway, I may as well remove it. Or I could add a `leaf` field to indicate which child locations are a leaf. What do you think?

sabrine33: thank you for looking into this: 
1) numStored property in the child,: it only gets populated correctly only the child is leaf, it contain actually some item, in case where it is a not leaf, the numStored = 0, so I could not use it for this purpose
2) Adding a leaf property my resolve the issue partly, as I may still need to do extra call is case it is not a leaf, 
or maybe I did not get the idea properly

khelwood: If the schema for a location would have
```graphql
Location {
    barcode
    ...
    """The other locations inside this location."""
    children: [
       { barcode: String!
         leaf: Boolean!
         numStored: Int!
         ...
       }, ...
    ]
}
```
then when you're loading the current location, you'd only have to see whether or not the `children` are leafs, and if they are, show the `numStored`. You wouldn't have to make requests for every child location. Does that make sense?

sabrine33: I think that will do, thank you David

khelwood: OK, I will add that. :+1:

khelwood (approved):

PR #803, opened 2024-11-21
set barcode information at the top right of the labware layout on the sectioning pages, labware details page, source and detination labware on visium transfer, cytassis and visium perm pages

sabrine33:


src/lib/helpers.ts
```diff
@@ -83,6 +83,12 @@ export enum GridDirection {
   /** Down the rightmost column, then left to the next column, etc. - Rotate 180  */
   LeftUp = 'LeftUp'
 }
+export enum Position {
+  TopRight = 'top right',
```
khelwood, line 87: Consistent with the others, the string would be `'TopRight'`.

khelwood (approved):

PR #861, opened 2025-01-15
X1264 section thickness per slot

sabrine33:


src/components/labware/Labware.tsx
```diff
@@ -342,7 +352,7 @@ const Labware = ({
               text={slotText}
               secondaryText={slotSecondaryText}
               color={_slotColor}
-              selected={selectedAddresses?.has(address)}
+              selected={selectedAddresses?.has(address) || (highlightedSlots?.has(address) ?? false)}
```
khelwood, line 355: Can you explain what this is saying for my own understanding please?

src/components/planning/Planner.tsx
```diff
@@ -220,10 +224,11 @@ export default function Planner<M>({
     dispatch({
       type: 'ADD_LABWARE_PLAN',
       labwareType: selectedLabwareType,
-      numLabwareAdd: numPlansToCreate ?? 1
+      numLabwareAdd: numPlansToCreate ?? 1,
+      sectionThickness: sectionThickness ?? 0.5
```
khelwood, line 228: Is this defaulting section thickness to 0.5? and is that the correct behaviour?

sabrine33: The value displayed by default in the UI is 0.5, if not updated, that is what we are sending to the core

khelwood (approved):

PR #917, opened 2025-03-26
Invalid timestamp in probe hybridisation QC

sabrine33:
The format_date_time function was called on time change, issue was occurring when the user sends the form with the loaded date-time (without updating it). Moving the format_date_time function to be called before sending the form

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "client",
-  "version": "3.4.4",
+  "version": "3.4.5",
   "private": true,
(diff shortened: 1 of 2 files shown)


khelwood (approved):
If you're confident about this, I'll put it straight out onto production.

sabrine33:
> If you're confident about this, I'll put it straight out onto production.

Yes, I think it is safe to do so - thanks

khelwood:
@sabrine33 OK, that's released.

PR #928, opened 2025-04-29
input fields in some forms are not quite visible

sabrine33:


khelwood (approved):

khelwood:
@sabrine33 When you link a pull request with an issue, the issue will be automatically closed when the pull request is merged, which makes it disappear from the board. Preferably don't do this.

PR #972, opened 2025-06-17
use Qrcode generator to generate 2D barcode

sabrine33:


khelwood:
Is it possible to generate a datamatrix instead of a qr code?

khelwood:
> > Is it possible to generate a datamatrix instead of a qr code?
> 
> It is possible, but there is no official out-of-the-box library to do so — so it would likely require a lot of workarounds. Unless there is a valid reason preventing us from using QR codes, I think it’s more reasonable to stick with QR codes.

I've sent a message to Katy to ask if their scanners can read a qr code.

Issue #975, opened 2025-06-18
x1317 client—As TS (Katy) I want some stain types disabled that are no longer in use, on the stain operation page


khelwood:
It looks like the client has some hard-coded stuff related to stain types

PR #977, opened 2025-06-18
x1277 remove hard-coded combined-stain-type

khelwood:
For #975


khelwood:
See also [x1317_disable_stain_types.sql](https://github.com/sanger/stan-sql/blob/devel/static/x1317_disable_stain_types.sql)

PR #1017, opened 2025-07-31
x1279:size range for library concentration operation

sabrine33:


khelwood (commented):
Looks fine if the failing tests are fixed

PR #1022, opened 2025-08-01
x1420: update release file options

sabrine33:
no major changes for the client side, just update the sdk to match the graphql schema change


sabrine33:
> I think a change is required in the client. When the user is presented with release options when performing a release, they need to be shown only the enabled options; but when they are selecting options for a release that already happened, they need to be shown all the options, so in that case the options should be loaded with `includeDisabled: true`.

Thanks, I thought they only can download from the history page.

PR #1189, opened 2026-04-30
fix block processing

sabrine33:
Closes #

- 

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -14,5 +14,7 @@ export default defineConfig({
     },
     baseUrl: "http://localhost:3000",
     testIsolation: false,
+    viewportWidth: 1280,
+    viewportHeight: 720,
   },
 });
--- a/cypress/e2e/pages/blockProcessing.cy.ts
+++ b/cypress/e2e/pages/blockProcessing.cy.ts
@@ -55,25 +55,19 @@ describe('Block Processing', () => {
     });
   });
   describe('Multi sample Proviasette block', () => {
-    context('when adding a Proviasette with 2 rows and 2 columns', () => {
+    context('when adding a Proviasette', () => {
       before(() => {
         selectOption('labwareType', 'Proviasette');
-        cy.findByTestId('selectedLabwareNumRows').type('{selectall}2');
-        cy.findByTestId('selectedLabwareNumColumns').type('{selectall}2');
         cy.findByText('+ Add Labware').click();
       });
       after(() => {
         cy.findAllByText('Delete Layout').first().click();
       });
-
-      it('display labware layout with 4 slots', () => {
-        cy.findAllByTestId('slot').should('have.length', 4);
-      });
-      describe('when adding different sources to differen slots', () => {
+      describe('when adding different sources to different slots', () => {
         before(() => {
           cy.findAllByText('Edit Layout').first().click();
           cy.findByRole('dialog').within(() => {
-            cy.findByText('STAN-113').click();
+            cy.findAllByText('STAN-113').first().click();
             cy.findByText('A1').click();
             cy.findByText('A2').click();
             cy.findAllByText('STAN-213').first().click();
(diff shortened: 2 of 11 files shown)


cypress/e2e/pages/blockProcessing.cy.ts
```diff
@@ -106,15 +81,9 @@ describe('Block Processing', () => {
...
-          cy.findAllByText('Edit Layout').first().click();
-          cy.findByRole('dialog').within(() => {
-            cy.findByText('STAN-113').click();
-            cy.findByText('A1').click();
-            cy.findByText('Done').click();
-          });
-        });
         it('should display STAN-113', () => {
+          editLayout(0, ['STAN-113'], ['A1']);
+          cy.wait(5000);
```
khelwood: Is this wait(5000) really necessary? If so, could you add a comment to the code explaining why?

khelwood (changes requested):

cypress/e2e/pages/blockProcessing.cy.ts
sabrine33: we added that for debugging - I will removed with the force click too

cypress.config.ts
```diff
@@ -14,5 +14,7 @@ export default defineConfig({
     },
     baseUrl: "http://localhost:3000",
     testIsolation: false,
```
sabrine33, line 16: increasing the cypress browser window for better visibility when failure

src/lib/factories/sampleFactory.ts
```diff
@@ -28,7 +28,7 @@ export const sampleFactory = Factory.define<Sample>(({ sequence, params, associa
 export const tissueFactory: Factory<Tissue> = Factory.define<Tissue>(({ params, associations }) => ({
   __typename: 'Tissue',
   externalName: params.externalName ?? `${faker.person.lastName()}${faker.string.numeric()}`,
-  replicate: params.replicate ?? String(_.random(10)),
+  replicate: params.replicate ?? String(_.random(1, 10)),
```
sabrine33, line 31: making sure mocked samples always have replicate number

cypress/e2e/pages/visiumQC.cy.ts
```diff
@@ -45,10 +45,10 @@ describe('Visium QC Page', () => {
   describe('On Visium QCType as Slide Processing', () => {
```
sabrine33, line 45: update barcode just to use a different type of labware (visium) instead of block as blocks now are 10*10.

cypress/e2e/pages/sectioningPlanning.cy.ts
```diff
@@ -461,7 +461,7 @@ function createLabware() {
   cy.findByText('+ Add Labware').click();
   cy.findByText('Edit Layout').click();
   cy.findByRole('dialog').within(() => {
-    cy.findByText('STAN-113').click();
+    cy.findAllByText('STAN-113').first().click();
```
sabrine33, line 464: as a block can have multiple samples now

cypress/e2e/pages/blockProcessing.cy.ts
```diff
@@ -36,67 +36,42 @@ describe('Block Processing', () => {
   describe('Source labware table', () => {
```
sabrine33, line 36: Moved the test setup from the `before` block into the `it` block because failures in before can fail silently in Cypress, and the retry mechanism is not applied there.

yoldas (approved):
Converts Map iterators e.g. plans.values() using Array.from to be able to to use array methods like map, filter, and flatMap (iterators do not have those methods).

Next, we may update the project nvmrc (v22.14.0) and we need to update the container image used in test.yml cypress/browsers:node18.12.0-chrome106-ff106 with supported versions and match them.
Node 18 is not supported. https://endoflife.date/nodejs

Correction: v22 is still supported.

PR #1194, opened 2026-05-07
x1439 client - [PR] Treatment type on the SGP management page to help breakdown the worfklows

yoldas:

- Add treatment type fragments, mutations, and update queries (core and client branches have slightly different timelines; hence a few cherry picks for client - SDK to be rebuilt again at merge)
- Add treatment types multi-select to Allocate a new SGP number in SGP Management page
- Add treatment types to the success message
- Add treatment types column to work allocation table in SGP Management page
- Show treatment types as either blue (enabled) or red (disabled) pills in the table cell
- Add treatment types to Records for SGP management spreadsheet download (enabled and disabled)
- Add treatment types tab to Configuration page

User feedback: Treatment types should appear after Work Type
(except configuration page tab; they are in alphabetical order)

**Not included: "Editable on the final SGP tracking page (later story)"**

Miscellaneous:

- Update several existing tests. The main update is clearing text before input. Discovered in debugging that tests are not well isolated and input fields accumulate entries across tests. The other main update is async. waitFor to avoid timeout errors. There are also some fixes for false-positives, particularly in releaseOptions tests.
- ~Update cypress container image to 22.14.0 . It is multi-browser version, because chrome-only versions are no longer provided for node 22+ . Chrome version 134 is used for tests.~ [Removed after code-review feedback]
- Fix the issue in Config page for Treatment Types only: after in-client-app navigation, the options set previously look reset.
- Add unit tests at component and page level with mocks.
- **No guards agains null or undefined treatment types; assumed empty array.**


.github/workflows/test.yml
```diff
@@ -15,9 +15,8 @@ jobs:
     name: Run cypress tests
     runs-on: ubuntu-latest
     container:
-      image: cypress/browsers:node18.12.0-chrome106-ff106
-
-      options: --user 1001
+      image: cypress/browsers:node-22.14.0-chrome-134.0.6998.165-1-ff-136.0.3-edge-134.0.3124.85-1
```
sabrine33: I would leave this for a separate PR and handle the Node upgrade there, including upgrading the build Node version as well as the github actions versions

At the moment, the tests are running with a different Node version from the one used to build the image

src/pages/Configuration.tsx
```diff
@@ -658,6 +660,37 @@ export default function Configuration() {
...
+            const enabled = typeof value === 'boolean' ? value : false;
+            return stanCore
+              .SetTreatmentTypeEnabled({
+                enabled,
+                name: entity.name
+              })
+              .then((res) => {
+                // revalidate route loader so the configuration page sees server-side updates
+                try {
+                  revalidator.revalidate();
```
sabrine33, line 681: do you mean the server response by 'the server-side updates'?

tests/unit/components/workAllocation/workAllocation.spec.tsx
```diff
@@ -0,0 +1,167 @@
...
+// Test the Treatment Types field in the WorkAllocation form, including
+// enabled/disabled options, selecting/unselecting options, displaying
+// selected options as chips in the UI, and submitting the form with the
+// selected treatment types to see the success message.
+
+// Use shared URL.createObjectURL mock util for jsdom
+mockCreateObjectURL();
+
+// Mock URL.createObjectURL for jsdom environment
+mockCreateObjectURL();
```
sabrine33: this is called twice

tests/unit/components/workAllocation/workAllocationTable.spec.tsx
```diff
@@ -0,0 +1,166 @@
...
+    const table = await screen.findByTestId('work-allocation-table');
+    // find the row for our SGP
+    const sgpCell = within(table).getByText('SGP-1');
+    const row = sgpCell.closest('tr') as HTMLTableRowElement | null;
+    expect(row).not.toBeNull();
+
+    // treatment type pills rendered in the treatment cell
+    // find pill spans anywhere in the row
+    const allSpans = row!.querySelectorAll('span');
+    const pillSpans = Array.from(allSpans).filter((p) => {
```
sabrine33: If you're using a component specifically for test assertions, the convention is to add a data-testid attribute so it can be reliably selected and tested. This makes it easier to target elements in tests without relying on CSS classes or DOM structure.

sabrine33 (changes requested):

.github/workflows/test.yml
yoldas: Test use the version from .nvmrc which is 22.14.0 and that matches this image version. 
I will still extract the change to another PR as you asked.

src/pages/Configuration.tsx
yoldas: The config change updates the page and backend; not the client state, hence the previous values are loaded again after navigation between tabs or pages. 

The fix above is making sure that the saved new values are loaded after navigation. This is done only for the treatment type tab.

.github/workflows/test.yml
yoldas: I have restored it but I think it was using node 18 for image and 22 for tests.

sabrine33: Yes, you're right — they were using different Node versions! I will put a story in the backlog to update the node version and the github actions all together. thanks alot!

sabrine33 (approved):

PR #1208, opened 2026-06-27
CI Fixes

yoldas:
## stan-client CI Fixes

I have done this work because my PR CI runs took 2 hours several times (depending on GitHub, multiple pushes, etc). In best cases, PRs in this repo took more than half an hour consistently. 

See CI Actions results, timing is average ~13 minutes; instead of ~35 minutes now. The following are the list of improvements:

a6219de5 — Bump workflow action versions. Upgrade actions/checkout, actions/setup-node, actions/upload-artifact, actions/download-artifact, and cypress-io/github-action to current major versions. Resolves Node 16/20 deprecation warnings from GitHub Actions runners.

181a3cec — Suppress serve request logs. Add --no-request-logging to the static-server command used by the Cypress jobs so test output stays focused on Cypress results instead of being drowned in per-asset HTTP logs.

1f333c33 — Parallelise Cypress tests across 4 matrix shards. Introduce a containers: [1, 2, 3, 4] matrix and wire up cypress-split via SPLIT / SPLIT_INDEX env vars. Cypress wall-clock for a full run drops from ~20 min single-runner to ~7–11 min across four parallel shards at the time of testing.

bb36f79d — Tune Cypress retry counts. 1 retry in CI (to absorb genuinely flaky network/animation timing), 0 retries locally (so we see real failures immediately instead of hiding them behind a quiet retry).

fbea0580 — Cancel superseded test runs on PRs. Add a concurrency block that cancels in-progress runs when a new commit is pushed to the same PR, while still queueing (not cancelling) runs on devel so every commit gets a status.

6b21d431 — Upgrade Cypress browsers container to cypress/browsers:22.14.0. Aligns the CI container's Node version with .nvmrc (Node 22.14.0) and brings Chrome up to a recent stable; eliminates an engines mismatch that was producing yarn warnings.

6ab68292 + 3091f6bb + b8bc218d — Experiment with shared node_modules artifact (reverted). Tried sharing the installed node_modules (plus the Cypress binary cache) from the build job to all four Cypress shards via an uploaded tarball, to avoid running yarn install four times. In practice the 1.77 GB artifact download was wildly inconsistent across shards (22 s on the fastest, 2 min 2 s on the slowest in the same run) and produced no net wall-clock win over per-job yarn install with setup-node's yarn cache. Reverted to keep the workflow simple and predictable.

ad8e0059 — Bump remaining actions to Node 24 runtime. Upgrade actions/download-artifact from v5 to v8, cypress-io/github-action from v6 to v7, and JamesIves/github-pages-deploy-action from 4.1.5 to v4. All three previous pins targeted Node 20, which the GitHub Actions runner now forces to run on Node 24 with a deprecation warning. The new majors run natively on Node 24, silencing the warning across test.yml and deploy-storybook.yml.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,12 +14,12 @@ jobs:
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
         name: Checkout
       - name: Enable Corepack
         run: corepack enable
       - name: Setup node
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v6
         with:
           node-version-file: '.nvmrc'
           cache: 'yarn'
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -4,7 +4,7 @@ import cypressSplit from 'cypress-split';
 export default defineConfig({
   video: false,
   defaultCommandTimeout: 15000,
-  retries: 2,
+  retries: { runMode: 1, openMode: 0 },
   numTestsKeptInMemory: 0,
   experimentalMemoryManagement: true,
   e2e: {
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -1,4 +1,5 @@
 import { defineConfig } from 'cypress';
+import cypressSplit from 'cypress-split';
 
 export default defineConfig({
   video: false,
@@ -7,10 +8,10 @@ export default defineConfig({
   numTestsKeptInMemory: 0,
   experimentalMemoryManagement: true,
   e2e: {
-    // We've imported your old cypress plugins here.
-    // You may want to clean this up later by importing these.
     setupNodeEvents(on, config) {
-      return require("./cypress/plugins/index.js")(on, config);
+      cypressSplit(on, config);
+      require('./cypress/plugins/index.js')(on, config);
+      return config;
     },
     baseUrl: "http://localhost:3000",
     testIsolation: false,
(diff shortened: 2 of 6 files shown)


cypress.config.ts
```diff
@@ -1,16 +1,17 @@
...
+  retries: { runMode: 1, openMode: 0 },
   numTestsKeptInMemory: 0,
   experimentalMemoryManagement: true,
   e2e: {
-    // We've imported your old cypress plugins here.
-    // You may want to clean this up later by importing these.
     setupNodeEvents(on, config) {
-      return require("./cypress/plugins/index.js")(on, config);
+      cypressSplit(on, config);
+      require('./cypress/plugins/index.js')(on, config);
```
BenTopping, line 13: Looks like that plugins index file is empty so it can be removed.

BenTopping (commented):
Looks sensible to me.
Might be overkill but it looks like you can feed cypress-split a set of timings for each test file so it can balance out the overall timing better. Should prevent multiple long-running tests in the same node

.github/workflows/build.yml
```diff
@@ -14,12 +14,12 @@ jobs:
...
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
         name: Checkout
       - name: Enable Corepack
         run: corepack enable
       - name: Setup node
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v6
```
StephenHulme, line 22: There are sanger versions of these actions available. They work identically, but can be updated in one central location for all our repos.

StephenHulme (approved):
Nice changes, looks good!

khelwood (approved):
Looks fine to me. Maybe do a merge-squash instead of a regular merge.

cypress.config.ts
yoldas: Yes, I have considered that but it is not harmful and it was already there.

.github/workflows/build.yml
yoldas: Yes, maybe in another PR.

PR #1209, opened 2026-06-30
x1561 - sectioning order

sabrine33:
Closes #

- 

### DIFF
--- a/src/components/sectioningConfirm/sectioningConfirm.machine.ts
+++ b/src/components/sectioningConfirm/sectioningConfirm.machine.ts
@@ -472,9 +472,14 @@ function autoFillSectionNumbers(layoutPlan: LayoutPlan, incrementFill: boolean, 
     let newSectionNum = 0;
     if (startNumbers && startNumbers.has(plan.source.labware.barcode) && incrementFill) {
       //Get the highest section number of the source labware for this section
-      newSectionNum = startNumbers.get(plan.source.labware.barcode)! + 1;
-      //Store the current highest section number so that it will be incremental for next section
-      startNumbers.set(plan.source.labware.barcode, newSectionNum);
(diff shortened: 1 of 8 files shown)


src/components/sectioningConfirm/sectioningConfirm.machine.ts
```diff
@@ -472,9 +472,14 @@ function autoFillSectionNumbers(layoutPlan: LayoutPlan, incrementFill: boolean,
     let newSectionNum = 0;
     if (startNumbers && startNumbers.has(plan.source.labware.barcode) && incrementFill) {
       //Get the highest section number of the source labware for this section
-      newSectionNum = startNumbers.get(plan.source.labware.barcode)! + 1;
-      //Store the current highest section number so that it will be incremental for next section
-      startNumbers.set(plan.source.labware.barcode, newSectionNum);
+      const highestSectionNumber = startNumbers.get(plan.source.labware.barcode)!;
+      if (plan.sectioningOrder) {
```
khelwood, line 476: So you've gone with a system that will translate gaps in sectioning order to gaps in the section numbers. I don't think it's what we agreed. Is there a reason it should be like this?

sabrine33: Here, the system calculates the section number by adding the sectioning order (set by the user on the planning page) to the highest section number of the source.  The section number can be modified by the user in the confirm page

khelwood: So if the highest section for the source is 2 and the sectioning orders for this labware are 5,6,7, it'll get sections 7,8,9. All the section numbers between 3 and 6 would fall into a gap.

sabrine33: yes - this is what I mentioned in the last standup- and mentioned that the user needs to put the sectioning order by source.

PR #1223, opened 2026-07-23
x1562 - [PR] Edit treatment types

### LINKED ISSUE #1220: x1562	As TS (Katy) I want "treatment type" to be editable in the SGP management table, in case end users change their mind on type of work
PR: https://github.com/sanger/stan-client/pull/1223

"(option 2) Table below can have an additional Edit link and it opens a dialog for individual edits, for example a modal

Image

Image

Alternatively, the editing can be positioned with a small dialog, similar to status edit.
Table column width does not need adjusting. - overlay a small dialog or use modal.
Keeps existing label colouring
Editing becomes two step, saving is done explicitly with Save button. Cancel is possible.
Once saved, it can hide the disabled option to be added again.
Selected items in the menu may not need colouring to show disabled types."

yoldas:
Story: https://github.com/sanger/stan-client/issues/1220

- Adds the ability to edit treatment types on a work row inline


yarn.lock
sabrine33, line 1: why the lock file is updated without updating the json file

src/components/workAllocation/WorkRow.tsx
```diff
@@ -228,6 +238,95 @@ export default function WorkRow({
...
+            <PinkButton
+              action={'tertiary'}
+              type="button"
+              data-testid={`${workNumber}-edit-treatment-types`}
+              onClick={() => {
+                setSelectedTreatmentTypes(treatmentTypeNames);
+                setIsEditingTreatmentTypes(true);
+              }}
+            >
+              Edit
```
sabrine33: Can you make the text more explicit here, for example : edit treatment types ?

sabrine33 (commented):
a couple of minor comments

yarn.lock
yoldas: We don't have an entry for caniuse-lite in package.json . `yarn why caniuse-lite` shows it is a transitive  dependency of react-scripts. Instead of version changes, there is a recommended way of doing it
https://github.com/browserslist/update-db#readme . Because our yarn version is older, I have executed `npx update-browserslist-db@latest` to update it as the test logs showed.

I have done that to make CI logs of this PR cleaner. I can remove it if you think it is an issue.

src/components/workAllocation/WorkRow.tsx
yoldas: I kept it as Edit because it is in the same cell. Edit Treatment Types took too much space. However, I have changed it now as you suggested, and attached screenshots before and after.

Screenshot 2026-07-27 at 13 21 28
Screenshot 2026-07-27 at 13 45 16

sabrine33: The column should be already quite wide any wide as the title width 'Treatment types' - I suppose it is fine

sabrine33 (approved):

PR #1228, opened 2026-07-29
restrict project pills to the column width

sabrine33:
Closes #

- 


khelwood (commented):
Can you describe the change here? I think the change we talked about was letting pills overflow the column so they wouldn't distort the table; but this pr says "restrict project pills to the column width", which sounds like the opposite

sabrine33:
> Can you describe the change here? I think the change we talked about was letting pills overflow the column so they wouldn't distort the table; but this pr says "restrict project pills to the column width", which sounds like the opposite

The pill overflow is being hidden, which restricts the content to the width of the parent column.

PR #1229, opened 2026-07-29
update the node version to latest LTS

sabrine33:
- update the node version to latest LTS [24.18.0](https://endoflife.date/nodejs)

- <s>update tsconfig file accordingly to : https://github.com/tsconfig/bases/</s>

- reverting the tsconfig.file file as that requires ts upgrade along other libraries too.  And setting the target to 'esnext' as that latest ECMAScript version.
`The "target" option in tsconfig.json tells the TypeScript compiler which version of JavaScript to compile your code down to.`


khelwood (approved):

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2015",
+    "target": "esnext",
```
yoldas: can we set this to a stable ECMAScript release? ES2024 or ES2025 would be a better setting here. I will check TypeScript version and update the comment.

sabrine33: TypeScript version needs to be updated it self - once that is done we can update this config file accordingly to the official doc : https://github.com/tsconfig/bases/ 
But yes we can use a more stable version, I set it up to 'esnext' as that is what is actually specified in the other field 'module,..'

yoldas: Why ESNext? ESNext = ES2025 (latest) + future language proposals . Can't we set both to ES2025 ?

tsconfig.json
sabrine33, line 1: Module does not support  es2024 (nor es2025) entry.  - so left it to esnext

> TypeScript hasn't defined a distinct module emit profile for those years because no new ECMAScript module-system features were introduced that would require one — es2022 already covers top-level await, and anything newer is folded into esnext.
> esnext is a moving target that always maps to the latest module features TypeScript supports, so it's accepted even when a specific year isn't. If/when new module-level syntax lands (e.g. import attributes stabilizing), the team may add an es2024/es2025 value — but until then, only the explicitly listed years are valid.

yoldas (approved):

PR #1239, opened 2026-08-06
support an array of plannedAction per slot

sabrine33:

* Updated `layoutPlan` to support an array of `plannedActions`. Previously, `plannedActions` was a map where the key was either the slot address (when the slot was not grouped into a section) or the section ID (when it was grouped).
* Added a `sectionGroupId` property to `plannedAction` to indicate whether a slot belongs to a section and, if so, which section.
* A slot can now have multiple plannedActions to support the case where all samples within a block are assigned to the same tube programmatically. Assigning different sources to the same slot is not permitted through the UI.
* Updated the `Labware` and `LabwarePerSection` components to support the new `plannedActions` structure.
* I considered merging both components (Labware and LabwarePerSection), but doing so broke the sectioning behaviour. In the LabwarePerSection component, selecting a slot within a section automatically selects all slots in that section. This is the intended behaviour for the Xenium Analyser operation, but not for the other operations.
* Updated the `labwareHandler` to generate the different labware types correctly. This change caused some tests tests to fail because the incorrect labware type was being displayed for certain operations.
* Updated the affected tests accordingly.


src/components/planning/LabwarePlan.tsx
```diff
@@ -261,33 +261,32 @@ const LabwarePlan = React.forwardRef<HTMLDivElement, LabwarePlanProps>(
...
-                          {Object.keys(current.context.layoutPlan.plannedActions).length === 0 && (
+                          {current.context.layoutPlan.plannedActions.length === 0 && (
                             <MutedText>
                               Please transfer samples to the slot before entering the section thickness.
                             </MutedText>
                           )}
-                          {Object.keys(current.context.layoutPlan.plannedActions).map((sectionGroupId) => (
-                            <div key={sectionGroupId} className="grid grid-cols-3 text-center">
+                          {(isTube(outputLabware.labwareType)
+                            ? current.context.layoutPlan.plannedActions.slice(0, 1)
```
sabrine33, line 270: why slicing here

src/components/sectioningConfirm/confirmLabware.machine.ts
```diff
@@ -189,9 +196,10 @@ export const createConfirmLabwareMachine = (
             return context;
           }
           return produce(context, (draft) => {
-            draft.layoutPlan.plannedActions[event.sectionGroupId].source.commentIds = event.commentIds.map(
-              (commentID) => Number(commentID)
-            );
+            const planned = findPlannedActionBySectionGroupId(draft.layoutPlan.plannedActions, event.sectionGroupId);
+            if (planned) {
```
sabrine33, line 211: what if a slot does not belong to a section?

src/components/sectioningConfirm/confirmLabware.machine.ts
```diff
@@ -223,7 +230,10 @@ export const createConfirmLabwareMachine = (
...
           }
           return produce(context, (draft) => {
-            const plannedAction = draft.layoutPlan.plannedActions[event.sectionGroupId];
+            const plannedAction = findPlannedActionBySectionGroupId(
+              draft.layoutPlan.plannedActions,
+              event.sectionGroupId
+            );
 
             if (plannedAction) {
               plannedAction.source.newSection = event.sectionNumber;
```
sabrine33: what if the slot does not belong to a section

src/components/sectioningConfirm/confirmLabware.machine.ts
```diff
@@ -235,16 +245,20 @@ export const createConfirmLabwareMachine = (
             return context;
           }
           return produce(context, (draft) => {
-            draft.layoutPlan.plannedActions[event.sectionGroupId].source.sampleThickness = event.thickness;
+            const plannedAction = findPlannedActionBySectionGroupId(
+              draft.layoutPlan.plannedActions,
+              event.sectionGroupId
+            );
+            if (plannedAction) plannedAction.source.sampleThickness = event.thickness;
```
sabrine33, line 256: what if the slot does not belong to a section

src/components/sectioningConfirm/confirmLabware.machine.ts
```diff
@@ -256,7 +270,7 @@ export const createConfirmLabwareMachine = (
             barcode: context.labware.barcode!,
             cancelled: context.cancelled,
             confirmSections: context.cancelled ? undefined : confirmSections,
-            addressComments: []
```
sabrine33, line 259: why not init the addressComments here

src/components/sectioningConfirm/ConfirmTubes.tsx
```diff
@@ -228,11 +227,11 @@ const TubeRow: React.FC<TubeRowProps> = ({
             <Input
               data-testid={`section-thickness-tube-${layoutPlan.destinationLabware.barcode}`}
               type="number"
-              value={layoutPlan.plannedActions['A1'].source.sampleThickness}
+              value={layoutPlan.plannedActions[0].source.sampleThickness}
```
sabrine33, line 230: For the tubes - we can have an array of plannedActions - not sure why selection the first one

src/components/sectioningConfirm/ConfirmTubes.tsx
```diff
@@ -241,12 +240,12 @@ const TubeRow: React.FC<TubeRowProps> = ({
             <Input
               data-testid={`sectionnumber-tube-${layoutPlan.destinationLabware.barcode}`}
               type="number"
-              value={sectionNumber}
+              value={cancelled ? '' : layoutPlan.plannedActions[0].source.newSection}
```
sabrine33, line 243: we can not have multiple labware within the same slot/tupe - but we can have one labware with multiple samples - plannedActions expected to be an array of those samples - not sure why selecting the first one

src/components/sectioningConfirm/sectioningConfirm.machine.ts
```diff
@@ -319,11 +319,11 @@ export function createSectioningConfirmMachine() {
...
-              const sectionGroupId = Object.keys(planInContext.plannedActions).find(
-                (sectionGroupId) => sectionGroupId === event.sectionGroupId
+              const plannedSection = planInContext.plannedActions.find(
+                (sectionDetails) => sectionDetails.sectionGroupId === event.sectionGroupId
               );
-              if (sectionGroupId) {
-                planInContext.plannedActions[sectionGroupId].source.newSection = event.sectionNumber;
+              if (plannedSection) {
+                plannedSection.source.newSection = event.sectionNumber;
               }
```
sabrine33, line 326: what happens if a slot does not belong to a section

src/components/sectioningConfirm/sectioningConfirm.machine.ts
```diff
@@ -533,12 +534,12 @@ function buildLayoutPlans(plans: Array<FindPlanDataQuery>, sourceLabwares: Array
           tissue: planned.source.samples[0].tissue
         };
         sources.push(source);
-        const sectionGroupId = group.length === 1 ? group[0] : index.toString();
-        plannedActions[sectionGroupId] = {
+        plannedActions.push({
+          sectionGroupId: planIndex + groupIndex,
```
sabrine33: this is will not be unique

src/lib/machines/layout/layoutMachineOptions.ts
```diff
@@ -90,54 +49,73 @@ export const machineOptions: InternalMachineImplementations<LayoutMachineImpleme
...
-
-        if (
-          !context.selected ||
-          (plannedActions[event.address] && context.selected.sampleId === plannedActions[event.address].source.sampleId)
-        ) {
-          draft.layoutPlan.plannedActions = deleteDestinationAddressFromGroup(
-            draft.layoutPlan.plannedActions,
-            event.address
+        if (!context.selected || slotPreviousSampleIds.includes(context.selected.sampleId)) {
+          return;
```
sabrine33, line 74: `slotPreviousSampleIds.includes(context.selected.sampleId)` what this means

src/components/planning/LabwarePlan.tsx
sabrine33: Even when a tube has multiple planned actions (i.e. multiple samples from the same labware), they are still assigned to the first slot (A1) and should therefore have the same section thickness. The slicing is based on the table being defined per section/slot, whereas for a tube, there is always a single slot.

Putting screenshot, so to know which display this is responsible for ( Section planning page):

Screenshot 2026-09-02 at 13 14 07

src/components/sectioningConfirm/confirmLabware.machine.ts
sabrine33: Updated this to check by addresses - so to cover both cases (when an address belongs to a section and when it does not)

src/components/sectioningConfirm/confirmLabware.machine.ts
sabrine33: Updated this to check by addresses - so to cover both cases (when an address belongs to a section and when it does not)

src/components/sectioningConfirm/ConfirmTubes.tsx
sabrine33: Selecting the first one as this is responsible for the sample thickness table display per slot/section. Even though a tube can have multiple plannedActions - they all should have the same thickness when sectioning

src/lib/machines/layout/layoutMachineOptions.ts
sabrine33: `slotPreviousSampleIds`  contains the list of sampleIds that are been moved to the selected destination slot - if slotPreviousSampleIds contains the sample id - return, as we don't need to assign it here.

src/components/labware/Labware.tsx
```diff
@@ -392,11 +392,11 @@ const Labware = ({
       sectionGroups = sectionGroupsBySample(labware as LabwareFlaggedFieldsFragment);
     }
     if (!sectionGroups) return result;
-    Object.entries(sectionGroups)
-      .filter(([, sectionDetails]) => sectionDetails.addresses.size > 1)
-      .forEach(([groupId, sectionDetails]) => {
+    sectionGroups
+      .filter((sectionDetails) => sectionDetails.addresses.size > 1)
```
yoldas: note: only keeps groups with more than one address

src/components/originalSampleProcessing/blockProcessing/BlockProcessingLabwarePlan.tsx
```diff
@@ -56,13 +56,15 @@ type BlockProcessingLabwarePlanProps = {
 function buildInitialLayoutPlan(
   sourceLabware: Array<LabwareFlaggedFieldsFragment>,
   sampleColors: Map<number, string>,
-  outputLabware: NewFlaggedLabwareLayout
+  outputLabware: NewFlaggedLabwareLayout,
+  operationType?: string
```
yoldas: question: callers of this function do not supply operationType currently. What is the use of the operationType parameter?
src/components/originalSampleProcessing/blockProcessing/BlockProcessingLabwarePlan.tsx 
src/components/originalSampleProcessing/potProcessing/PotProcessingLabwarePlan.tsx
src/components/planning/LabwarePlan.tsx

src/components/planning/labwarePlan.machine.ts
```diff
@@ -236,7 +236,14 @@ export const createLabwarePlanMachine = (initialLayoutPlan: LayoutPlan) =>
           if (event.type !== 'UPDATE_SOURCES') {
             return context;
           }
-          const sources = convertLabwareTypeToSourceType(event.sources, event.sectionThickness?.toString());
+
+          const sources = convertLabwareTypeToSourceType(
+            event.sources,
+            event.sectionThickness?.toString(),
+            isPlanningByLabware(context.layoutPlan.destinationLabware.labwareType, context.layoutPlan.operationType)
```
yoldas: note: This change here looks good, but when operationType is undefined, the code always uses SourceUniqueBy.SAMPLE . This is related to the previous question about not supplying the operationType.

src/components/sectioningConfirm/confirmLabware.machine.ts
```diff
@@ -256,7 +273,7 @@ export const createConfirmLabwareMachine = (
...
```
yoldas: note: I think the previous version was a bug and the change fixed it.

yoldas (approved):

